PR #9, opened 2026-03-03
[OLD][HYPER-109] feat: custom CSS injection for trusted OAuth clients

aspiers:
## Summary

- Extract `resolveClientMetadata` and `ClientMetadata` type into `@certified-app/shared` so both pds-core and auth-service can use them
- Add `branding.css` support to `ClientMetadata` with `escapeCss()` (prevents `</style>` injection) and `getClientCss()` (trust-gated helper)
- **Auth-service**: inject trusted client CSS into login and consent pages via `<style>` tag (CSP already allows `'unsafe-inline'`)
- **PDS-core**: middleware intercepts `/oauth/authorize` responses — wraps `res.setHeader`/`res.end` to inject `<style>` tag and append SHA256 hash to CSP `style-src` directive
- **Demo app**: add sample branding CSS (gradient buttons, CSS custom properties) to client metadata
- Add `PDS_OAUTH_TRUSTED_CLIENTS` env var (comma-separated client_id URLs) to all env examples

## How it works

1. Trusted clients list their CSS in `client-metadata.json` under `branding.css`
2. Both pds-core and auth-service read `PDS_OAUTH_TRUSTED_CLIENTS` (must be set identically)
3. When a trusted client initiates OAuth, the CSS is fetched, escaped, and injected into authorization pages
4. PDS side: CSP hash is computed and appended to `style-src` since the stock oauth-provider uses strict hash-based CSP
5. Auth-service side: no CSP changes needed (`'unsafe-inline'` already allowed)

## Test plan

- [x] All 173 existing tests pass
- [x] TypeScript builds cleanly for shared, auth-service, and pds-core
- [x] Manual: start with `PDS_OAUTH_TRUSTED_CLIENTS` set to demo's client_id, verify CSS appears on login/consent pages
- [x] Manual: remove from trusted list, verify CSS no longer injected
- [x] Manual: verify CSP header includes CSS hash on PDS authorization page

Ref: atproto-9fa

## Summary by CodeRabbit

* **New Features**
  * Trusted OAuth clients can inject client-specific styling into consent, login and recovery pages.
  * OTP UI now adapts to configurable length/format, updating input attributes and user-facing descriptions.

* **Chores**
  * .env examples expanded with commented variables for trusted clients and OTP configuration.
  * Tests and demo metadata updated to include trusted client handling.


Kzoeps:
@aspiers updated the logic to fetch the custom css a bit (refer to this [commit](https://github.com/hypercerts-org/ePDS/pull/9/changes/976e6709de86793d9393fd035e2f8b310f5efcc2))

we now pass the actual `request_uri` in the query params like how bsky oauth does it. previously it was passing a placeholder. We then use that `request_uri` to fetch the client id and subsequently the client-metadata if the query params is missing the client-id.

aspiers:
Superseded by #48

PR #11, opened 2026-03-11
fix: require consent on account creation

Kzoeps:
Previously the condition to show the consent page was `if !newAccount` but this doesnt make sense since its still good to display what perms the app is asking.

Also the stock pds also displays the consent page even if theyre creating a new account

## Summary by CodeRabbit

* **Refactor**
  * Simplified consent requirement logic for client authentication flows

### DIFF
--- a/packages/pds-core/src/index.ts
+++ b/packages/pds-core/src/index.ts
@@ -65,7 +65,6 @@ async function main() {
     const requestUri = req.query.request_uri as string
     const email = ((req.query.email as string) || '').toLowerCase()
     const approved = req.query.approved === '1'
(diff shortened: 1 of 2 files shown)


packages/pds-core/src/index.ts
```diff
@@ -65,7 +65,6 @@ async function main() {
     const requestUri = req.query.request_uri as string
     const email = ((req.query.email as string) || '').toLowerCase()
     const approved = req.query.approved === '1'
-    const _isNewAccount = req.query.new_account === '1'
```
Kzoeps: unused code

PR #29, opened 2026-03-17
make handle selection on signup optional (issue #22)

### LINKED ISSUE #22: Make handle selection on signup optional

Kzoeps:
## Summary
- Adds `HANDLE_SELECTION_MODE` env var to control whether users can choose their handle during signup
- Supports `user-choice`, `random`, and `none` modes, with fallback precedence chain
- Fixes premature auth_flow cleanup and step label issues in random-mode flow

Closes #22

Also closes HYPER-192 on linear

## Summary by CodeRabbit

* **New Features**
  * Multiple signup handle modes: random, picker, picker-with-random.
  * Random signup skips the picker and proceeds directly; picker-with-random shows a "Generate random handle" button with availability checks, retries, and timeouts; button is disabled during submission.

* **Tests**
  * Expanded coverage for handle-mode resolution, persistence, client-metadata fallback, PAR ping, and callback signature contract.

* **Documentation**
  * Updated env example with EPDS_DEFAULT_HANDLE_MODE and session-secret guidance.

### DIFF
--- a/packages/auth-service/src/routes/choose-handle.ts
+++ b/packages/auth-service/src/routes/choose-handle.ts
@@ -574,8 +574,10 @@ function renderChooseHandlePage(
       });
 
       // Random handle button: generate a base36 local part client-side (mirrors
-      // generateRandomHandle() in shared/src/crypto.ts) and confirm availability
-      // via /api/check-handle, retrying up to 3 times on collision.
+      // generateRandomHandle() in shared/src/crypto.ts — duplicated here because
+      // this script is inlined in a template literal and cannot import server-side
+      // modules) and confirm availability via /api/check-handle, retrying up to
+      // 3 times on collision.
       var randomBtn = document.getElementById('random-btn');
       if (randomBtn) {
         function randomLocalPart() {


.env.example
```diff
@@ -68,20 +66,22 @@ PDS_ADMIN_PASSWORD=
 PDS_PORT=3000
 PDS_DATA_DIRECTORY=/data
 
-PDS_DID_PLC_URL=https://plc.directory
-PDS_BSKY_APP_VIEW_URL=https://api.bsky.app
-PDS_BSKY_APP_VIEW_DID=did:web:api.bsky.app
-PDS_CRAWLERS=https://bsky.network
-
+# Generate with: openssl rand -hex 32
```
aspiers: Another deleted comment and some reordering which I'm not sure adds any benefit and creates headaches when comparing the `.env` files of existing deployments with `.env.example` to figure out how to migrate them to the latest codebase.

If a commit is going to reorder things, it needs to explain why, and not mix the reordering with other edits otherwise it's a big headache to review.

aspiers: Same fix — commit 618c541 restores all removed comments and reverts the reordering. The diff vs main is now clean: only the `AUTH_SESSION_SECRET` hint and the new `EPDS_DEFAULT_HANDLE_MODE` block remain.

packages/auth-service/src/routes/login-page.ts
```diff
@@ -56,6 +89,12 @@ export function createLoginPageRouter(ctx: AuthServiceContext): Router {
       return
     }
 
+    const clientMeta = await safeResolveClientMetadata(clientId)
+    const handleMode = resolveHandleMode(
+      req.query.epds_handle_mode as string | undefined,
+      clientMeta,
+    )
```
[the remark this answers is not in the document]
aspiers: Small correction to ~my~ (**EDIT:** *AI's*) previous reply: Kzoeps didn't claim the RFC rules it out — just that the library doesn't support it. The only correction needed is to the "doesn't make sense" framing: RFC 9126 explicitly allows custom/extension parameters in PAR, so it would make sense protocol-wise. The library is the sole blocker.

aspiers (approved):

PR #32, opened 2026-03-19
feat: handle dead/drifted accounts on /account with sign-out and error page

Kzoeps:
## Summary
When a user signs into /account with a valid better-auth session but the underlying PDS account no longer exists for that email (e.g. the account was deleted out-of-band, or the email was changed via XRPC without going through /account), the previous code had no handling for this case. The user would land on the account settings page with no DID — seeing handle update and backup email features that silently fail — or get caught in a redirect loop with no explanation.

## Changes

This PR makes that failure mode explicit and recoverable:
- getDidByEmail now returns a typed union ({ did: string } | { did: null } | null) so call sites can distinguish a confirmed "no account" from a PDS outage — previously both collapsed to the same code path, which caused returning users to incorrectly receive a "Welcome" OTP email during a PDS blip.
- /account now detects the no-account case, signs out the dangling session, and shows a clear error page explaining what happened — breaking the redirect loop and giving the user actionable guidance instead of leaving them stuck.
- requireBetterAuth was extracted from account-settings.ts into middleware/ (alongside csrf.ts and rate-limit.ts) so it could be properly unit-tested — all 5 branches are now covered in require-better-auth.test.ts.

## Summary by CodeRabbit

* **New Features**
  * Added middleware to validate sessions, populate request-local session and account ID, and shared HTML renderers for error/no-account pages.

* **Bug Fixes**
  * Distinguish lookup failures (now surface 503 and abort OTP) from “no account” (403); flows and responses updated accordingly.

* **Tests**
  * Added/expanded tests covering middleware branches, OTP/account-lookup cases, and HTML renderers.

* **Documentation**
  * Updated testing/coverage notes to reflect improved middleware unit-test coverage.


Kzoeps:
think we arent going to be doing it this way anyways so closing this PR

PR #46, opened 2026-04-06
fix: disable buttons once approve/deny clicked on consent

Kzoeps:
Bug: Double-clicking "Approve" or "Deny" on the consent form submitted two POST requests to /auth/consent. The first POST successfully read the auth_flow row and then immediately deleted it. By the time the second POST arrived (milliseconds later), the row was gone — causing getAuthFlow to return undefined and the user to see "Authentication session expired. Please try again."

Fix: Added onsubmit="this.querySelectorAll('button').forEach(b => b.disabled = true)" to the consent form. The moment either button is clicked, both buttons are disabled in the browser before the form submits, making it physically impossible to send a second POST.

swappy-20260406-124334

## Summary by CodeRabbit

* **Bug Fixes**
  * Consent form now disables form buttons on submission to prevent duplicate submissions and improve reliability and user experience.

* **Tests**
  * End-to-end test environment updated with OTP configuration (length and charset) to better mirror production OTP behavior during CI test runs.


aspiers:
No longer applicable since we switched away from this broken consent screen to the stock PDS consent screen.

PR #114, opened 2026-04-27
Skip chooser when login_hint targets an unbound account (Flow 1)

aspiers:
## Summary

- New Flow 1 gate: when `login_hint` resolves to an email NOT bound to the current device, auth-service skips the chooser redirect and renders the email/OTP form for a fresh sign-in. Previously it would either auto-select the wrong account (single-binding device) or surface the hinted user's mailbox to a stranger (multi-binding device).
- Cookies are left intact — only this single request bypasses session reuse, so other accounts on the device remain reusable on subsequent un-hinted visits.
- Hint-less flows are unchanged.

### Plumbing

- `pds-core`: new `/_internal/device-accounts` endpoint behind `EPDS_INTERNAL_SECRET` returns the lowercased emails bound to a `(dev-id, ses-id)` pair, validated against the device row before listing accounts. Lookup factored into a shared `loadDeviceAccountEmails` helper.
- `auth-service`: `shouldReuseSession` accepts an optional `SessionReuseHintContext`; new `readDeviceSessionCookies` extracts cookie values for the bindings fetch; `login-page` hoists hint resolution above the session-reuse decision and only fetches bindings when both a hint and a cookie pair are present (no extra round trip on hint-less or cookie-less flows).

### Tests

- 14+ new unit tests in `session-reuse.test.ts` covering match / mismatch / stale-pair / legacy / case-insensitive paths.
- New `device-accounts.test.ts` (pds-core) and `fetch-device-accounts.test.ts` (auth-service) for the helpers.
- New BDD scenario for the hint-matches-binding path in `session-reuse-bugs.feature`.
- Hint-mismatch BDD scenario added as `@pending` — it needs a second-account fixture (separate browser context, account creation, teardown). Unit tests cover the decision logic exhaustively.

## Test plan

- [x] `pnpm format:check`
- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm test` (752/752 passing)
- [x] `pnpm docker:build`
- [x] Cucumber dry-run resolves all new steps
- [ ] Manual: trigger Flow 1 with a stranger email as `login_hint` against the running stack and confirm the OTP form renders
- [ ] Add a second-account e2e fixture and unmark the `@pending` mismatch scenario in a follow-up

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Enhanced session reuse with login hints: when a login hint (email) is provided, the system now checks device-binding status. Bound accounts show the account chooser (pre-selected), while unbound accounts require OTP verification and disable session reuse for that request.
  * Direct routing to email verification form when redirected with pre-filled identity information not matching existing signed-in accounts.

* **Tests**
  * Added comprehensive end-to-end and unit tests for session reuse behavior with login hints.


.changeset/cross-client-session-reuse.md
```diff
@@ -13,10 +13,12 @@ Signing in once in your browser now works across all apps that use this ePDS.
 - A "Use a different account" link on the chooser takes you back to the email form for a fresh sign-in.
 - The chooser shows your email next to your handle so accounts are easy to tell apart.
 - If your browser's leftover sign-in cookies no longer match the server, you land on the familiar email code form rather than a generic sign-in screen.
+- If an app sends you to ePDS asking for a specific email or handle that is not one of the accounts you have already signed in to in this browser, you go straight to the email code form for that account rather than landing on a chooser of your existing accounts.
```
aspiers: No, the fix is flow 1 where the *app* asks for your *email*.  And this is the ePDS so we don't generally ask for handles!

.changeset/cross-client-session-reuse.md
```diff
@@ -13,10 +13,12 @@ Signing in once in your browser now works across all apps that use this ePDS.
...
 
 **Client app developers:** no client-side changes required.
 
 - When a previous sign-in's cookies are present, the user lands on the account chooser to confirm which identity to reuse.
+- When you set `login_hint` to an email, AT Protocol handle, or DID, ePDS checks whether the hinted account is bound to the current device. If it is, the chooser still appears (or auto-selects the only matching account). If not, session reuse is disabled for this single request and the user receives an OTP for the hinted account; other accounts on the device remain reusable on subsequent un-hinted visits — no cookies are cleared.
```
aspiers: Why did you write it auto-selects the only account?  Pretty sure this doesn't happen!

features/session-reuse-bugs.feature
```diff
@@ -103,3 +103,28 @@ Feature: Session-reuse resilience against stale device cookies
...
+  # is not bound to the current device, auth-service must skip session reuse
+  # and surface the email/OTP form so the hinted user can sign in fresh.
+  # See packages/auth-service/src/lib/session-reuse.ts shouldReuseSession
+  # and packages/pds-core/src/lib/device-accounts.ts.
+  # ---------------------------------------------------------------------------
+
+  Scenario: Login hint matches a bound account — chooser still wins
+    When the demo client starts a new OAuth flow with the test user's handle as login_hint
+    Then the browser lands on the ePDS enriched account picker
+    And the stock upstream welcome page is not shown
```
aspiers: Isn't this redundant if the previous `Then` succeeds? The page can't be both the account picker *and* the welcome page?

features/session-reuse-bugs.feature
```diff
@@ -103,3 +103,28 @@ Feature: Session-reuse resilience against stale device cookies
...
+
+  @pending
+  # Needs a second isolated account whose handle is registered on the PDS
+  # but not bound to the current device. The fixture lift (separate browser
+  # context, account creation, then teardown) is non-trivial and out of
+  # scope for the patch that introduces the gate; unit tests in
+  # session-reuse.test.ts cover the decision logic exhaustively.
```
aspiers: This is absolutely in scope and needs to be done correctly in this PR!

features/session-reuse-bugs.feature
```diff
@@ -103,3 +103,28 @@ Feature: Session-reuse resilience against stale device cookies
...
+  # ---------------------------------------------------------------------------
+  # Flow 1 hint-vs-bindings gate: when login_hint resolves to an email that
+  # is not bound to the current device, auth-service must skip session reuse
+  # and surface the email/OTP form so the hinted user can sign in fresh.
+  # See packages/auth-service/src/lib/session-reuse.ts shouldReuseSession
+  # and packages/pds-core/src/lib/device-accounts.ts.
+  # ---------------------------------------------------------------------------
+
+  Scenario: Login hint matches a bound account — chooser still wins
+    When the demo client starts a new OAuth flow with the test user's handle as login_hint
```
aspiers, line 107: How does this ensure the account is already bound?

packages/auth-service/src/lib/session-reuse.ts
```diff
@@ -95,12 +123,51 @@ export function isForceLoginPrompt(req: SessionReuseRequest): boolean {
...
+  const resolvedEmail = hintCtx?.resolvedEmail
+  if (!resolvedEmail) return true
+  // Hint present but caller didn't supply a bindings list (legacy call
+  // sites): preserve existing behaviour rather than silently disabling
+  // reuse.
+  if (hintCtx.deviceBoundEmails === undefined) return true
+  // Hint present and caller could not resolve a bindings list (cookie
+  // pair was malformed or stale at pds-core): treat as no usable session.
+  if (hintCtx.deviceBoundEmails === null) return false
+  return hintCtx.deviceBoundEmails.includes(resolvedEmail.toLowerCase())
```
aspiers: Is `hintCtx.deviceBoundEmails` guaranteed to be lowercase? Seems brittle

.changeset/cross-client-session-reuse.md
aspiers: Fixed in 8b1dfc1 — reworded as `If an app sends you to ePDS pre-filled with an email that is not one of the accounts...`. Dropped the "or handle" misdirection.

.changeset/cross-client-session-reuse.md
aspiers: You're right. Upstream's auto-SSO branch (oauth-provider.js:321) needs `prompt == null`, but ePDS forces `prompt='consent'` for unauthenticated clients (pds-core/src/index.ts:484), so the chooser always renders. Reworded in 8b1dfc1 to say the chooser appears with the hinted account pre-selected.

features/session-reuse-bugs.feature
aspiers: Agreed — redundant. Dropped from this scenario in 8b1dfc1. (The other scenarios in the file have the same redundancy but I left those alone, out of scope for this PR.)

features/session-reuse-bugs.feature
aspiers: Done in 8b1dfc1. Removed `@pending`, added a working second-account fixture: new `Given another user has a separate PDS account` creates the second user inside an isolated browser context (so its cookies don't bleed onto the primary device), captures email/handle/did onto new world fields, then closes that context. New `Then the OTP form is pre-filled with the other user's email` asserts on the hidden `#otp-email` value to prove the resolved hint reached OTP. Both Login-hint scenarios pass against the docker stack:

```
2 scenarios (2 passed)
14 steps (14 passed)
```

features/session-reuse-bugs.feature
aspiers: Implicit via Background. `a returning user has a PDS account` runs `createAccountViaOAuth` which goes through pds-core's `/oauth/epds-callback`, which calls `provider.accountManager.upsertDeviceAccount(deviceId, account.sub)` (packages/pds-core/src/index.ts:385). Then `the user has completed one OAuth sign-in from the demo client` re-signs in (also via the OAuth flow → another `upsertDeviceAccount`). So by the time this When fires, the test user IS bound. Happy to add an explicit `And the test user is bound to the device` Given that hits `/_internal/device-accounts` if you'd prefer the assertion be visible, but it's belt-and-braces over what the Background already guarantees — let me know.

packages/auth-service/src/lib/session-reuse.ts
aspiers: Fair point. Fixed in f3fdc00 — re-normalise both sides at comparison time so the gate's correctness doesn't depend on producer discipline:

```ts
const target = resolvedEmail.toLowerCase()
return hintCtx.deviceBoundEmails.some((e) => e.toLowerCase() === target)
```

.changeset/cross-client-session-reuse.md
aspiers: Still wrong! There is no pre-filling here! As I just said, the app asks the user for their email. It's flow 1. Please don't make me repeat myself yet again.

features/session-reuse-bugs.feature
aspiers: They are not out of scope. If you don't fix them now, they will get forgotten.

features/session-reuse-bugs.feature
aspiers: OK that's fine

e2e/step-definitions/session-reuse-bugs.steps.ts
```diff
@@ -171,6 +172,119 @@ When(
...
+    // skip session reuse and surface the email/OTP form.
+    const page = getPage(this)
+    const url = new URL('/api/oauth/login', testEnv.demoUrl)
+    url.searchParams.set('login_hint', this.otherUserHandle)
+    await page.goto(url.toString())
+  },
+)
+
+Then(
+  "the OTP form is pre-filled with the other user's email",
```
aspiers: "pre-filled" makes it sound like there is a field the user could type into, but you're actually talking about a hidden field here, so fix the misleading text.

.changeset/cross-client-session-reuse.md
aspiers: Reworded again in 1b14351 to centre the app-asks-for-email framing: `If an app asks you for your email and you give it one that is not one of the accounts you have already signed in to in this browser, you go straight to the email code form for that account...`

features/session-reuse-bugs.feature
aspiers: You're right. Done in 1b14351 — dropped `And the stock upstream welcome page is not shown` from every scenario in the file (10 occurrences) and removed the now-unused step definition. Each preceding `Then` already pins host + a host-unique element (chooser hydration data on pds-core, `#email` on auth-service) so the welcome-page case is precluded.

e2e/step-definitions/session-reuse-bugs.steps.ts
aspiers: Right. Renamed in 539c0de to `the OTP form will submit the other user's email` so the step name describes what's actually asserted (the hidden `#otp-email` value rides along on the submit). Comment in the step body now also calls out that it's a hidden input.

features/session-reuse-bugs.feature
```diff
@@ -70,7 +64,6 @@ Feature: Session-reuse resilience against stale device cookies
     Then the browser lands on the ePDS enriched account picker
     When the user clicks "Another account" on the enriched account picker
     Then the browser lands on the auth-service email-and-OTP form
-    And the stock upstream welcome page is not shown
     And the upstream stock sign-in form is not shown
```
aspiers: Isn't this redundant too?

packages/pds-core/src/lib/device-accounts.ts
```diff
@@ -0,0 +1,83 @@
+/**
+ * Device-bound account lookup shared between welcome-page-guard and the
```
aspiers, line 2: I don't get it - this is a brand new file but there are no changes to welcome-page-guard?

features/session-reuse-bugs.feature
aspiers: Yes, redundant by the same logic. Dropped in e3c7ddb (this scenario plus the matching one in passwordless-authentication.feature:150 that uses the same step). Removed the unused step definition too.

packages/pds-core/src/lib/device-accounts.ts
aspiers: Good catch — file's header comment claimed the helper was "shared between welcome-page-guard and the /_internal/device-accounts endpoint" but the guard kept its own inline copy. Refactored in e0261a9: the guard now calls `loadDeviceAccountEmails` and bounces on `null || emails.length === 0` (covering parse-fail / stale-pair / no-bindings via a single check). Drops ~50 lines of duplicate validate-cookie-then-list logic. All 37 guard unit tests still green.

PR #117, opened 2026-04-28
Fix host-only cookie shadowing OTP loop (#116)

### LINKED ISSUE #116: OTP login loops back to OTP screen when browser holds stale host-only device cookies
## Summary

After a user submits a valid OTP, they are redirected straight back to the OTP screen instead of landing in the client app. The flow loops indefinitely. Reproduced today on `epds1.test.certified.app` with a user whose browser had stale host-only `dev-id`/`ses-id` cookies left over from a session predating the cookie-domain broadening change. Clearing cookies for both `epds1.test.certified.app` and `auth.epds1.test.certified.app`, or running in incognito, works around it.

Two bugs combine to produce the loop. Either alone would not be enough to break the user; both together turn a recoverable cookie state into an unrecoverable trap. A separate architectural concern is also worth flagging.

## Expected flow (existing user, ePDS OAuth)

1. App: user clicks "Sign in" → browser redirected to **auth-service** `/oauth/authorize`.
2. **auth-service**: checks for an existing session via cookies. None usable → renders the email form.
3. Browser: user types email, submits.
4. **auth-service**: emails an OTP, renders the OTP form.
5. Browser: user enters OTP, submits.
6. **auth-service** `/auth/complete`: verifies OTP. Redirects browser to **pds-core** `/oauth/epds-callback`.
7. **pds-core** `/oauth/epds-callback`: creates/loads a device session (sets fresh `dev-id` + `ses-id` cookies on the response), binds the account to that device. Redirects browser to **pds-core** `/oauth/authorize`.
8. Browser: follows the redirect, sending its cookies back to pds-core.
9. **pds-core** safety check (welcome-page-guard) on `/oauth/authorize`: re-reads `dev-id` + `ses-id` from the request. If valid → passes through to upstream OAuth provider. If not → bounces back to auth-service `/oauth/authorize?prompt=login` and tries to clear the bad cookies.
10. **pds-core** upstream provider (when the guard passes): renders the consent screen, or auto-approves and issues an authorization code, then redirects browser back to the app with the code.
11. App: exchanges the code for tokens. Done.

## Actual flow when bug fires

Steps 1–8 run normally. At step 9, the guard bounces. The bounce sends the user back to step 2, which renders the OTP form (step 4). The user re-enters the OTP. Steps 5 → 6 → 7 → 8 → 9 repeat. Forever.

## Bug 1 — Stale host-only cookies parsed first when no Domain-scoped pair exists

After PR #103, pds-core's cookie-domain middleware rewrites every outbound `Set-Cookie` for `dev-id`/`ses-id` to add `Domain=<pds-host>`. This is necessary so auth-service (on a sibling subdomain) can read the cookies for cross-client session reuse.

However, browsers treat host-only and Domain-scoped cookies with the same name as **two distinct entries** in the cookie jar. A user whose jar held a host-only pair from before PR #103 is in trouble whenever the host-only entry is what the cookie parser ends up validating.

The precise condition: the `cookie` package's `parse()` (used by both upstream's `parseHttpCookies` and welcome-page-guard's `parseDeviceCookies`) keeps the **first** occurrence of each name when duplicates exist in the request's Cookie header. Per RFC 6265 §5.4, browsers serialise cookies into the Cookie header by path-length descending, then by **creation time ascending** for ties. Both pairs have `Path=/`, so creation time decides the order:

- A user who **only** has the host-only pair (no Domain-scoped pair has ever been set) — every request validates against host-only stale values. Loop fires.
- A user with a host-only pair created **before** any Domain-scoped pair (Ken's case: the host-only entry pre-dates PR #103, the Domain-scoped pair was created later by a fresh sign-in attempt that didn't complete) — host-only comes first, validates against stale values. Loop fires.
- A user with a host-only pair created **after** a Domain-scoped pair (e.g. someone who manually plants a synthetic host-only entry today, on top of an existing valid Domain-scoped pair) — Domain-scoped comes first, the parser picks the legitimate values, the guard passes. Loop does NOT fire.

So the loop affects users who carried a host-only pair across PR #103 without ever reaching a state where a fresh Domain-scoped pair gets created earlier.

## Bug 2 — The guard's host-only cookie clear is itself rewritten to Domain-scoped

The bounce in step 9 calls `appendCookieClearHeaders` ([`packages/pds-core/src/welcome-page-guard.ts:158`](packages/pds-core/src/welcome-page-guard.ts#L158)) which is meant to evict cookies in **both** scopes:

```ts
res.append('Set-Cookie', `${name}=; Max-Age=0; Path=/`)        // intended: host-only clear
if (cookieDomain) {
  res.append('Set-Cookie', `${name}=; Max-Age=0; Path=/; Domain=${cookieDomain}`)
}
```

Browsers only clear a host-only cookie if the clearing `Set-Cookie` itself has **no** `Domain=` attribute. The first line above is meant to do exactly that.

But the cookie-domain middleware ([`packages/pds-core/src/cookie-domain.ts:71`](packages/pds-core/src/cookie-domain.ts#L71)) sees a `Set-Cookie` for `dev-id`/`ses-id` with no `Domain=` and unconditionally appends one. The "host-only clear" is rewritten into a "Domain-scoped clear" that's now identical to the second line.

Result: the response carries two Domain-scoped clears and zero host-only clears. The host-only `dev-id`/`ses-id` survive every bounce. Verified in the captured response headers — every `dev-id=`/`ses-id=` clear in the response has `Domain=epds1.test.certified.app`.

So the loop has no termination: each bounce attempts to clean up the stale state but the cleanup itself is silently neutered.

## Reproduction (manual)

In DevTools → Application → Cookies on `epds1.test.certified.app`:

1. Delete any existing `dev-id` and `ses-id` rows (both scopes — host-only and Domain-scoped).
2. Add a `dev-id` row with **Domain blank** (i.e. host-only on the pds-core host), Path `/`, value `dev-` followed by 32 lowercase hex chars (e.g. `dev-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa`). Mark HttpOnly + Secure.
3. Add a `ses-id` row the same way (e.g. `ses-bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb`).
4. Make sure **no other** `dev-id`/`ses-id` rows exist on this host.
5. Visit the demo and try to log in. You should loop on the OTP form.

If you have a Domain-scoped `dev-id`/`ses-id` pair in the jar that was created *before* the planted host-only entries, the loop will not reproduce — that pair comes first in the Cookie header and the parser picks it. Either delete the Domain-scoped pair or do this in a fresh browser profile.

## Proposed fixes

The proper fix is two narrow changes that together resolve the stale-cookie loop without restructuring the flow:

1. **Fix the rewrite bug (Bug 2)**: `rewriteSetCookie` in [`cookie-domain.ts`](packages/pds-core/src/cookie-domain.ts) must not add `Domain=` to clearing cookies (`Max-Age=0` or past `Expires`). Detect and pass through unchanged. Add unit test covering the host-only-clear case. This restores the welcome-page-guard's ability to actually clear stale host-only cookies on a bounce.

2. **Proactively clear host-only twins on every fresh device-session Set-Cookie (Bug 1)**: in `/oauth/epds-callback` (the one place where we know we just minted a fresh device session), the response that sets the new Domain-scoped `dev-id`/`ses-id` should also explicitly clear any host-only `dev-id`/`ses-id` (and `:hash` sidecars). The host-only clear must bypass the cookie-domain rewrite (or rely on the Bug 2 fix to pass it through unchanged) so it actually evicts the stale entries. With this in place, no user reaches step 8 with a host-only entry parsed ahead of the fresh Domain-scoped pair, even if their jar was already polluted before the fix shipped.

Both changes are testable in isolation. Bug 2 is a unit-test fix in the rewriter. Bug 1's fix is exercised by an e2e scenario: start with only a host-only pair on the jar, run an OAuth flow with OTP submit, assert the user reaches the demo's `/welcome` page rather than looping on OTP.

## Architectural concern (separate, do not bundle)

The welcome-page-guard runs on **every** GET to `/oauth/authorize` and `/account*` regardless of where the request came from — post-OTP hop, session-reuse path, page refresh, browser back-button, prefetch, stale URL pastes, future entry points. Every scenario where the request's cookie state isn't pristine becomes an OTP request, regardless of whether re-authenticating actually makes sense for that scenario.

The guard exists for one specific purpose — preventing upstream from rendering its stock welcome page — but as a side effect it now gates every `/oauth/authorize` GET. Any future cookie-handling change, parser quirk, or new entry point can become an infinite OTP loop the same way the present bug did.

A less invasive alternative would be intercepting upstream's response and replacing the welcome page with a redirect, rather than pre-empting the request. That keeps upstream's normal session-handling behavior on healthy flows and only reacts when upstream actually decided to render the page we want to hide. Worth re-evaluating, but separately from the present fix — the two narrow changes above don't depend on the guard's broader role.

## References

- [`packages/pds-core/src/welcome-page-guard.ts`](packages/pds-core/src/welcome-page-guard.ts) — the guard, `appendCookieClearHeaders`, `buildBounceUrl`
- [`packages/pds-core/src/cookie-domain.ts`](packages/pds-core/src/cookie-domain.ts) — `rewriteSetCookie`, the middleware that breaks the host-only clear
- [`packages/pds-core/src/index.ts`](packages/pds-core/src/index.ts) — `/oauth/epds-callback` handler, redirects to `/oauth/authorize`
- [`packages/auth-service/src/lib/session-reuse.ts`](packages/auth-service/src/lib/session-reuse.ts) — `shouldReuseSession`, the `prompt=login` honouring branch the bounce relies on
- PR #103 — introduced the cookie-domain broadening

aspiers:
## Summary

Fixes the infinite OTP loop diagnosed in #116. A user whose browser jar carries a stale host-only `dev-id`/`ses-id` pair from before PR #103 (cookie-domain broadening) gets trapped on the OTP form: every OAuth attempt completes the OTP step, redirects through `/oauth/epds-callback`, then gets bounced from `/oauth/authorize` back to OTP because the welcome-page-guard validates against the stale host-only values instead of the freshly-set Domain-scoped pair.

Two narrow changes resolve it; either alone is insufficient.

## Bug 2 — `cookie-domain` middleware destroys host-only cookie clears

`rewriteSetCookie` rewrote every device-cookie `Set-Cookie` to add `Domain=<parent>`, including `Max-Age=0` clears. The welcome-page-guard's bounce response intentionally emits TWO clears per cookie name (one host-only, no `Domain=`; one Domain-scoped) so browsers evict cookies in both scopes. The middleware silently rewrote the host-only line to be Domain-scoped, identical to the second line, so the host-only stale entry survived every bounce.

**Fix:** `rewriteSetCookie` now skips clearing cookies (detected by `Max-Age=0` or past `Expires`). Normal `Set-Cookie` values continue to get `Domain=` injected. Unit tests added for each device cookie name, whitespace tolerance, past `Expires`, and the no-double-inject guard.

## Bug 1 — Stale host-only entry parsed first, shadowing the freshly-set Domain-scoped pair

When a request carries both pairs in the Cookie header, the `cookie` package's `parse()` (used by both upstream's `parseHttpCookies` and welcome-page-guard's `parseDeviceCookies`) keeps the first occurrence per name. Per RFC 6265 §5.4 ordering, the host-only stale entry comes first when its creation time predates the Domain-scoped pair (Ken's case — host-only existed pre-PR-#103). The guard validates the wrong values, fails, bounces.

**Fix:** in `/oauth/epds-callback`, the response that emits the fresh Domain-scoped cookies also emits explicit host-only `Max-Age=0` clears for `dev-id`, `ses-id`, and their `:hash` sidecars. The browser evicts any host-only twin before the very next request, so the welcome-page-guard at `/oauth/authorize` sees only the fresh pair. Idempotent — clearing cookies that don't exist is a no-op. Relies on Bug 2's fix to pass the host-only clears through unchanged.

## Tests

- New unit tests in `cookie-domain.test.ts` for the rewriter's clearing-cookie behaviour (Max-Age=0 host-only clears pass through unchanged for each device cookie name; positive Max-Age and future Expires still get `Domain=` injected; whitespace tolerated).
- New e2e scenario in `features/session-reuse-host-only-shadowing.feature`: an affected user (only host-only stale pair on the jar, no Domain-scoped, no live device session) starts an OAuth flow and submits a valid OTP. They must reach the demo client's `/welcome` page rather than loop.
- Full session-reuse e2e profile: 16/16 passing.

## Changeset

No new changeset. This is a fix for an unreleased feature (PR #103, `.changeset/cross-client-session-reuse.md` is unreleased). The existing changeset's user-facing claim ("If your browser's leftover sign-in cookies no longer match the server, you land on the familiar email code form rather than a generic sign-in screen") still holds post-fix — the bug was that the recovery path looped infinitely instead of recovering, not that the recovery existed at all.

## Architectural follow-up (NOT in this PR)

The welcome-page-guard runs on every GET to `/oauth/authorize` and `/account*`, so any future cookie-handling regression or new entry point can become an OTP loop the same way this bug did. Issue #116 flags a less invasive alternative (intercept upstream's welcome-page response and replace it, rather than pre-empting every request) for separate evaluation.

## Test plan

- [x] `pnpm vitest run packages/pds-core/src/__tests__/cookie-domain.test.ts` — 33/33 passing
- [x] `pnpm test:e2e:headless -p session-reuse` — 16/16 passing against local docker stack
- [x] Manual repro per #116 instructions: plant host-only `dev-id`/`ses-id` on `epds-poc1.test.certified.app`, log in via demo, reach `/welcome` without looping
- [ ] CI green
- [ ] Manual repro on `epds1.test.certified.app` with Ken's account once deployed

Closes #116

## Release Notes

* **Bug Fixes**
  * Fixed a critical issue where returning users with stale session cookies would experience repeated sign-in prompts. Users can now successfully complete authentication and access the application.

* **Tests**
  * Added comprehensive end-to-end scenario tests validating sign-in recovery behavior.
  * Added detailed unit test coverage for session cookie handling and eviction logic.


e2e/step-definitions/session-reuse-bugs.steps.ts
```diff
@@ -478,3 +478,108 @@ When(
...
+    ).toBe(pdsHost)
+    expect(
+      sesEntries[0].domain,
+      `ses-id should be host-only on ${pdsHost} but Playwright reports domain=${sesEntries[0].domain}`,
+    ).toBe(pdsHost)
+  },
+)
+
+When(
+  'the user submits a valid OTP for the existing account',
```
aspiers: Surely we can reuse existing steps here?

features/session-reuse-host-only-shadowing.feature
aspiers, line 1: Couldn't this have gone in `session-reuse-bugs.feature`?

e2e/step-definitions/session-reuse-bugs.steps.ts
aspiers: Done in 820af82 — replaced the bespoke step with the existing four-step sequence (`enters the test email`, `shows an OTP verification form`, `an OTP email arrives in the mail trap`, `enters the OTP code`). The custom step is gone.

features/session-reuse-host-only-shadowing.feature
aspiers: Done in 9a1dcda — merged the scenario into `session-reuse-bugs.feature`. The standalone file is gone. To make the Feature-level Background work for an affected-user starting state, the `browser jar holds only a stale host-only dev-id and ses-id pair` Given now clears the jar before planting, so it overrides whatever the Background's OAuth sign-in deposited.

PR #129, opened 2026-04-30
Welcome-page guard: cover upstream sign-in-view (rows 5/6/9)

### LINKED ISSUE #131: Welcome-page guard misses sign-in-view leaks (rows 5/6/9)
## Summary

The pre-route welcome-page guard at `packages/pds-core/src/welcome-page-guard.ts` only handles upstream's stock **welcome page** (rows 1–4). It misses upstream's **sign-in-view** (handle + password form), which renders whenever every binding upstream considers has `loginRequired: true` (rows 5, 6, 9).

ePDS accounts are passwordless (random unguessable password generated at creation, never exposed). Letting upstream render the password form is therefore a contract violation — it must never be reachable from any flow.

### Inputs per scenario

| \# | Scenario | Cookies | Bindings | PAR prompt | login_hint | Auth age > 7d? |
| -- | -- | -- | -- | -- | -- | -- |
| 1 | No cookies | absent | n/a | any | any | n/a |
| 2 | Orphan / malformed | partial | n/a | any | any | n/a |
| 3 | Stale (ses-id mismatch) | valid format, no DB match | n/a | any | any | n/a |
| 4 | Valid, zero bindings | valid + match | 0 | any | any | n/a |
| 5 | Valid, ≥1 binding, forced login | valid + match | ≥1 | login | any | any |
| 6 | Valid, ≥1 binding, all stale | valid + match | ≥1 | not login | any | true for all |
| 7 | Valid, ≥1 fresh binding, no hint OR hint→fresh | valid + match | ≥1 | not login | absent OR matches a fresh binding | false for ≥1 (incl. hinted) |
| 8 | prompt=none, no SSO match | valid + match | ≥1 | none | any | n/a |
| 9 | hint→stale binding, others fresh | valid + match | ≥1 | not login | matches one binding, that one stale | true for hinted; false for ≥1 other |

### Upstream behaviour + guard decision

| \# | Upstream would render | Guard today | Guard proposed |
| -- | -- | -- | -- |
| 1 | welcome | bounce | bounce |
| 2 | welcome | bounce | bounce |
| 3 | welcome | bounce | bounce |
| 4 | welcome | bounce | bounce |
| 5 | sign-in-view | pass (leak) | bounce |
| 6 | sign-in-view | pass (leak) | bounce |
| 7 | chooser / consent / SSO redirect | pass | pass |
| 8 | error redirect to client (no UI) | pass | pass |
| 9 | chooser with hinted account selected + loginRequired → user click → sign-in-view | pass (leak) | bounce |

**Note on row 9.** Technically the chooser renders first, then sign-in-view only after a click. Whether that counts as a "leak" depends on whether you treat the chooser as safe when it pre-selects an account that can only proceed via password. I'd argue yes — the user-visible outcome is "I clicked my account and got asked for a password", which is the same leak. Bouncing matches the contract more cleanly.

## Bounce rule

The bounce rule for the guard simplifies to: **bounce if every binding that could satisfy this request is** `loginRequired === true` — where "could satisfy" means matches `login_hint` (if set) or any binding (if no hint). Combined with `prompt === 'login'` forcing all-`loginRequired` upstream, that one rule covers rows 5, 6, and 9.

## Repro

Reproduced live against `epds1.test.certified.app`: a returning user clicked sign-in on the demo client and landed on the upstream password form, with no obvious recovery path back to OTP.

E2E scenarios for rows 5/6/9 are in `features/session-reuse-bugs.feature` under the "Sign-in-view leaks" section, runnable via `pnpm test:e2e:headless -p session-reuse` against a docker-compose stack. Rows 6 and 9 use a new `EPDS_TEST_HOOKS=1`-gated `/_internal/test/expire-device-account` route to backdate `account_device.updatedAt` past upstream's 7-day limit without waiting wall-clock.

## Follow-up

hypercerts-org/ePDS#133 — replace the pre-emptive logic with response-rewrite detection (less brittle than mirroring upstream's render decision).

aspiers:
## Summary

The pre-route welcome-page guard in pds-core today only intercepts upstream's stock welcome page (rows 1–4 of the failure-mode taxonomy in `docs/design/session-reuse-bugs.md`). It misses upstream's other authentication UI — its sign-in-view (handle + password form) — which renders whenever every binding upstream considers has `loginRequired: true`. ePDS accounts are passwordless, so any path into that form is unusable.

This PR adds e2e coverage for the three independent triggers, plus a pds-core test hook needed to drive rows 6 and 9. **The guard fix itself is the next commit on this branch** (PR is opened as draft; will be marked ready once the contract goes from red to green).

## Triggers

See #131 for table which lists various sign-in cases requiring guards.

- **Row 5** — stored PAR `parameters.prompt === 'login'`. Upstream marks every session `loginRequired` and renders sign-in-view directly.
- **Row 6** — every binding's `account_device.updated_at` is older than upstream's `authenticationMaxAge` (7 days). Upstream renders the chooser pre-selected on the stale account; clicking it falls through to sign-in-view.
- **Row 9** — `login_hint` resolves to a binding that's individually stale, even though other bindings on the same device are fresh. Chooser pre-selects the hinted (stale) account; click → sign-in-view.

## Reproductions captured

All three scenarios fail today at the `Then the browser lands on the auth-service email-and-OTP form` assertion, with screenshots showing upstream's sign-in or chooser UI on `pds-core`.

## Test hook

A new `/_internal/test/expire-device-account` route on pds-core backdates `account_device.updated_at` past upstream's 7-day max age. Mounted only when `EPDS_TEST_HOOKS=1`, refuses to start if `NODE_ENV=production`, authenticated via the existing `x-internal-secret` header. Mirrors auth-service's `/_internal/test/expire-otp` pattern.

## Drive-by fix

The pre-existing returning-user step had `page.fill('#code', otp)` against the auth-service login page's hidden `#code` input — broken because hidden inputs aren't fillable. Replaced with the existing `fillOtp` helper that fills the visible per-digit `.otp-box` inputs (which feed the hidden field via JS and auto-submit on the last digit).

## Test plan

- [x] Row 5 reproduces on a fresh docker-compose stack (lands on sign-in-view)
- [x] Row 6 reproduces on a fresh docker-compose stack (lands on chooser with single stale account)
- [x] Row 9 reproduces on a fresh docker-compose stack (lands on chooser with both accounts; hinted one is stale)
- [ ] Guard fix lands so all three scenarios go green
- [ ] Existing rows 1–4 + chooser/consent/auto-SSO scenarios continue to pass

Fixes #131.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session reuse detection for stale device bindings and aged account bindings.
  * Enhanced authentication to properly handle re-authentication requirements in multi-account device scenarios.
  * Refined OTP-based authentication flows for account recovery workflows.

* **Tests**
  * Expanded end-to-end test coverage for session reuse edge cases and authentication flow variations.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,6 +8,13 @@ services:
     env_file: .env
     environment:
       - PDS_PORT=3000
+      # Enable /_internal/test/* hooks used by the e2e suite (e.g.
+      # backdating account_device.updatedAt past upstream's 7-day
+      # authenticationMaxAge so checkLoginRequired returns true). Mounted
+      # only here and on the Railway pr-base preview env; production
+      # deployments leave it unset. The router throws at startup if
+      # NODE_ENV=production.
+      - EPDS_TEST_HOOKS=${EPDS_TEST_HOOKS:-1}
     volumes:
       - pds-data:/data
     restart: unless-stopped
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,13 +8,14 @@ services:
     env_file: .env
     environment:
       - PDS_PORT=3000
-      # Enable /_internal/test/* hooks used by the e2e suite (e.g.
+      # Optional /_internal/test/* hooks used by the e2e suite (e.g.
       # backdating account_device.updatedAt past upstream's 7-day
-      # authenticationMaxAge so checkLoginRequired returns true). Mounted
-      # only here and on the Railway pr-base preview env; production
-      # deployments leave it unset. The router throws at startup if
-      # NODE_ENV=production.
-      - EPDS_TEST_HOOKS=${EPDS_TEST_HOOKS:-1}
+      # authenticationMaxAge so checkLoginRequired returns true).
+      # Defaults OFF so the local docker-compose stack matches a
+      # production deployment by default; opt in for e2e runs by setting
+      # EPDS_TEST_HOOKS=1 in your .env or shell. The router additionally
+      # throws at startup if NODE_ENV=production, regardless of this var.
+      - EPDS_TEST_HOOKS=${EPDS_TEST_HOOKS:-0}
     volumes:
       - pds-data:/data
     restart: unless-stopped
@@ -34,11 +35,12 @@ services:
     env_file: .env
     environment:
       - AUTH_PORT=3001
-      # Enable /_internal/test/* hooks used by the e2e suite (e.g. forcing
-      # OTP expiry without a 10-minute wait). Mounted only here and on the
-      # Railway pr-base preview env; production deployments leave it unset.
-      # The router throws at startup if NODE_ENV=production.
-      - EPDS_TEST_HOOKS=${EPDS_TEST_HOOKS:-1}
+      # Optional /_internal/test/* hooks used by the e2e suite (e.g. forcing
+      # OTP expiry without a 10-minute wait). Defaults OFF — opt in for e2e
+      # runs by setting EPDS_TEST_HOOKS=1 in your .env or shell. The router
+      # additionally throws at startup if NODE_ENV=production, regardless
+      # of this var.
+      - EPDS_TEST_HOOKS=${EPDS_TEST_HOOKS:-0}
     volumes:
       - auth-data:/data
     restart: unless-stopped
--- a/e2e/step-definitions/session-reuse-bugs.steps.ts
+++ b/e2e/step-definitions/session-reuse-bugs.steps.ts
@@ -22,7 +22,7 @@ import { expect } from '@playwright/test'
 import type { EpdsWorld } from '../support/world.js'
 import { testEnv } from '../support/env.js'
 import { getPage } from '../support/utils.js'
-import { createAccountViaOAuth } from '../support/flows.js'
+import { createAccountViaOAuth, pickHandle } from '../support/flows.js'
 import { sharedBrowser } from '../support/hooks.js'
 import { clearMailpit, extractOtp, waitForEmail } from '../support/mailpit.js'
 import { fillOtp } from '../support/otp.js'
@@ -652,7 +652,7 @@ async function expireDeviceAccount(opts: {
     method: 'POST',
     headers: {
       'Content-Type': 'application/json',
-      'X-EPDS-Internal-Secret': testEnv.internalSecret,
+      'x-internal-secret': testEnv.internalSecret,
     },
     body: JSON.stringify(opts),
   })
@@ -698,6 +698,7 @@ Given('the device has two bound accounts', async function (this: EpdsWorld) {
   const message = await waitForEmail(`to:${otherEmail}`)
   const otp = await extractOtp(message.ID)
   await fillOtp(page, otp)
+  await pickHandle(this)
   await page.waitForURL('**/welcome', { timeout: 30_000 })
 
   const bodyText = await page.locator('body').innerText()
--- a/e2e/step-definitions/session-reuse-bugs.steps.ts
+++ b/e2e/step-definitions/session-reuse-bugs.steps.ts
@@ -625,23 +625,6 @@ When(
   },
 )
 
-Then(
-  'no upstream password field is rendered anywhere on the page',
-  async function (this: EpdsWorld) {
-    // Negative assertion: ePDS accounts are passwordless, so a password
-    // input on any screen we land on is a leak. Upstream's sign-in-view
-    // renders <input type="password">; auth-service's OTP form does not.
-    // Counting password inputs across the whole document keeps the
-    // assertion robust against future structural changes.
-    const page = getPage(this)
-    const passwordCount = await page.locator('input[type="password"]').count()
-    expect(
-      passwordCount,
-      `Found ${passwordCount} password input(s) on ${page.url()} — ePDS is passwordless, this is the upstream sign-in-view leak`,
-    ).toBe(0)
-  },
-)
-
 // Backdate `account_device.updated_at` past upstream's authenticationMaxAge
 // (7d) via a pds-core /_internal/test hook gated on EPDS_TEST_HOOKS=1. The
 // hook accepts a target DID and optional deviceId; omitting deviceId
--- a/e2e/step-definitions/session-reuse-bugs.steps.ts
+++ b/e2e/step-definitions/session-reuse-bugs.steps.ts
@@ -196,6 +196,29 @@ When(
   },
 )
 
+When(
+  "the demo client starts a new OAuth flow with the test user's handle as a PAR-body login_hint",
+  async function (this: EpdsWorld) {
+    if (!this.userHandle) {
+      throw new Error(
+        'world.userHandle missing — Background step must run first',
+      )
+    }
+    // login_hint_location=body puts the hint in the PAR body rather than the
+    // /oauth/authorize redirect URL — this matches the third-party
+    // OAuth-client pattern that drives upstream's matchesHint logic against
+    // `parameters.login_hint` from the stored PAR. Without this, upstream's
+    // request-manager-loaded parameters carry no hint and the guard's row 9
+    // bounce condition can never fire (because filterCandidateBindings sees
+    // every binding as a candidate, not just the stale one).
+    const page = getPage(this)
+    const url = new URL('/api/oauth/login', testEnv.demoUrl)
+    url.searchParams.set('login_hint', this.userHandle)
+    url.searchParams.set('login_hint_location', 'body')
+    await page.goto(url.toString())
+  },
+)
+
 Given(
   'another user has a separate PDS account',
   async function (this: EpdsWorld) {
--- a/e2e/step-definitions/session-reuse-bugs.steps.ts
+++ b/e2e/step-definitions/session-reuse-bugs.steps.ts
@@ -335,7 +335,15 @@ Then(
   'the browser lands on the auth-service email-and-OTP form',
   async function (this: EpdsWorld) {
     const page = getPage(this)
-    await expect(page.locator('#email')).toBeVisible({ timeout: 30_000 })
+    // Wait for one of the auth-service login-page steps to be visible.
+    // initialStep is 'email' on no-hint flows and 'otp' when a login_hint
+    // resolves to an email — both land on the same login-page route, just
+    // with a different step displayed initially. Either is a valid "lands
+    // on the login page" outcome; the assertion is "we're on auth-service,
+    // not pds-core".
+    await expect(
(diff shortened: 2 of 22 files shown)


e2e/step-definitions/session-reuse-bugs.steps.ts
```diff
@@ -564,3 +567,171 @@ Given(
...
+    // pds-core epds-callback redirects back to /oauth/authorize, the
+    // stored PAR still carries `prompt=login` — the row-5 trigger.
+    const page = getPage(this)
+    const url = new URL('/api/oauth/login', testEnv.demoUrl)
+    url.searchParams.set('prompt', 'login')
+    await page.goto(url.toString())
+  },
+)
+
+When('the user completes the OTP flow', async function (this: EpdsWorld) {
```
aspiers: This is worded too vaguely - what does "completes" mean exactly?

e2e/step-definitions/session-reuse-bugs.steps.ts
```diff
@@ -564,3 +567,171 @@ Given(
...
+  await expect(page.locator('#step-otp.active')).toBeVisible({
+    timeout: 30_000,
+  })
+  const message = await waitForEmail(`to:${this.testEmail}`)
+  const otp = await extractOtp(message.ID)
+  await fillOtp(page, otp)
+})
+
+Then(
+  'no upstream password field is rendered anywhere on the page',
```
aspiers: Don't we already have / need a step for this in other scenarios?

packages/pds-core/src/index.ts
```diff
@@ -1088,6 +1089,83 @@ async function main() {
...
+  // and refused outright when NODE_ENV=production. Mirrors auth-service's
+  // /_internal/test/expire-otp pattern: a narrow UPDATE that backdates a
+  // single timestamp to reproduce time-dependent behaviour without waiting
+  // out the wall-clock TTL. Used by features/session-reuse-bugs.feature
+  // rows 6 and 9 to age account_device.updatedAt past upstream's
+  // authenticationMaxAge (7 days) so checkLoginRequired returns true for
+  // the targeted binding(s).
+  // =========================================================================
+
+  if (process.env.EPDS_TEST_HOOKS === '1') {
```
aspiers: This is bloating main() which is already too big.

packages/pds-core/src/auth-ui-guard.ts
```diff
@@ -229,18 +232,167 @@ export function createWelcomePageGuard(opts: {
...
       logger,
     })
-    if (!emails || emails.length === 0) {
+    if (!bindings || bindings.length === 0) {
+      bounceOrPass()
+      return
+    }
+
+    // At this point bindings exist, so upstream won't render the welcome
+    // page. But it may still render its sign-in-view (handle + password
```
aspiers, line 258: Doesn't this mean that this guard is now misnamed, if it's guarding against two pages not just the welcome page?

packages/pds-core/src/auth-ui-guard.ts
```diff
@@ -229,18 +232,167 @@ export function createWelcomePageGuard(opts: {
...
+    if (params?.prompt === 'login') {
+      bounceOrPass()
+      return
+    }
+    const candidates = filterCandidateBindings(bindings, params?.login_hint)
+    if (candidates.every((b) => provider.checkLoginRequired(b))) {
       bounceOrPass()
       return
     }
     next()
```
aspiers, line 291: In what circumstances do we want to let requests through?  If it's going to render other pages?  Are we implementing guards by trying to replicate the same logic which is internal to the upstream PDS code for determining which page to render?  Isn't that very brittle?  Ideally we could just catch it exactly when the upstream has decided it wants to render the welcome page or the login page.  I'm guessing you have gone for this brittle approach because we don't have access to that internal PDS logic, is that right?

e2e/step-definitions/session-reuse-bugs.steps.ts
aspiers: Fair — the step was a monolithic catch-all. Fixed in 5961183 by replacing it in row 5 with the canonical decomposed steps already used across the suite (`enters the test email` → `OTP email arrives` → `enters the OTP code`), and dropped the bespoke step body entirely.

e2e/step-definitions/session-reuse-bugs.steps.ts
aspiers: Yes, the canonical sequence (`enters the test email` / `OTP email arrives` / `enters the OTP code`) was already in use across the suite. Fixed in 5961183 — row 5 now composes the existing steps and the bespoke `completes the OTP flow` step body is gone.

packages/pds-core/src/index.ts
aspiers: Agreed. Extracted in 5961183 to a new `packages/pds-core/src/lib/test-hooks.ts` (matches the `lib/` pattern used for the other extracted middleware). main() now calls a single `installTestHooks({ pds, app, logger })`.

packages/pds-core/src/auth-ui-guard.ts
aspiers: Agreed the name no longer captures the full scope. Two options:

1. **Rename file + function** to something like `upstream-auth-ui-guard` / `createUpstreamAuthUiGuard`. More accurate but churns every import + the welcome-page-guard.test.ts file + every reference in the docstring of `isGuardedPath` etc.
2. **Keep names, expand the file-level docstring** to make the expanded scope explicit and explain the historical naming.

In 5961183 I went with option 2 — names retained, top docstring rewritten to describe both UIs (welcome page rows 1–4; sign-in-view rows 5/6/9) with a sentence noting the names are kept for stability. Happy to do (1) if you'd rather pay the rename cost — let me know and I'll push a follow-up.

packages/pds-core/src/auth-ui-guard.ts
aspiers: Yes, your guess is right — that is exactly why we ended up with this approach, and yes it is brittle. Some context on the tradeoff:

**The pass-through cases.** With bindings present and at least one fresh + (no `prompt=login` in the stored PAR) + (`login_hint` either absent or matching a fresh binding), upstream renders the **enriched account chooser**, which IS reachable from ePDS by design. So pass-through is correct here, not a fallback. The guard fires only when upstream's next render would be one of the two unreachable UIs (welcome page or sign-in-view).

**What we'd ideally do.** Intercept upstream's rendering decision *itself* — wrap the response and check the rendered HTML for the welcome-page or sign-in-view markers (e.g. the React hydration JSON, the page title, an input `type=password`), then redirect if matched. That's the exact same pattern we already use in `createChooserEnrichmentMiddleware` (response-rewrite for `/account*`) and `createUpstreamFaviconMiddleware`. **Catch the page upstream actually decided to render, instead of pre-empting that decision.** That avoids the brittleness entirely.

**Why not in this PR.** Two reasons. First, response-rewrite means a 200 OK with HTML is in flight before we redirect — so we'd need to swallow the response body and emit a fresh 303, which works but is more invasive than a pre-route guard. Second, the existing `createWelcomePageGuard` already pre-empts the welcome-page case using the same kind of upstream-internal logic (zero bindings → welcome page), so this PR extends the existing pattern rather than rewriting it. A response-rewrite refactor would replace BOTH the welcome-page check and the sign-in-view check, which is a separate piece of work.

I'll file a follow-up issue to track the response-rewrite refactor as a more durable replacement for both checks. Worth doing — let me know if you'd rather block this PR on it.

aspiers: Filed: #133

packages/pds-core/src/auth-ui-guard.ts
aspiers: Renamed in ac6b8c7: `welcome-page-guard.ts` → `auth-ui-guard.ts`, `createWelcomePageGuard` → `createAuthUiGuard`. Updated all import sites + cross-package comments + design docs. The single remaining mention of the old name is an example string in `.agents/skills/writing-changesets/SKILL.md` (a representative example for changeset-writing prose; out of scope for this PR).

packages/pds-core/src/auth-ui-guard.ts
```diff
@@ -229,18 +240,129 @@ export function createWelcomePageGuard(opts: {
...
+      requestUrl: req.url,
       logger,
     })
-    if (!emails || emails.length === 0) {
+    if (params?.prompt === 'login') {
+      bounceOrPass()
+      return
+    }
+    const candidates = filterCandidateBindings(bindings, params?.login_hint)
+    if (candidates.every((b) => provider.checkLoginRequired(b))) {
```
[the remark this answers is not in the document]
aspiers: @coderabbitai Where is it documented that space-delimited sets are supported?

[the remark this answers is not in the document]
aspiers: @coderabbitai But we are not using OIDC here, are we?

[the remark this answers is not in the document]
aspiers: Good catch — fixed. Added `parsePromptTokens()` and `promptHasLogin()` helpers in `auth-ui-guard.ts` and switched both call sites (the guard's bounce condition and the epds-callback's strip) to tokenise. New unit tests cover the multi-token bounce, OIDC §3.1.2.1 spec citation in the comments. The strip in `index.ts` now keeps non-`login` tokens intact (a hypothetical `prompt="login consent"` becomes `prompt="consent"` after the OTP cycle, not silently dropped).

docker-compose.yml
```diff
@@ -8,6 +8,13 @@ services:
     env_file: .env
     environment:
       - PDS_PORT=3000
+      # Enable /_internal/test/* hooks used by the e2e suite (e.g.
+      # backdating account_device.updatedAt past upstream's 7-day
+      # authenticationMaxAge so checkLoginRequired returns true). Mounted
+      # only here and on the Railway pr-base preview env; production
+      # deployments leave it unset. The router throws at startup if
+      # NODE_ENV=production.
+      - EPDS_TEST_HOOKS=${EPDS_TEST_HOOKS:-1}
```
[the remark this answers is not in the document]
aspiers: Agreed — flipped both `core` and `auth` defaults to `${EPDS_TEST_HOOKS:-0}`. The local docker-compose stack now matches a production deployment by default; opt in for e2e runs by setting `EPDS_TEST_HOOKS=1` in your `.env` or shell. The `NODE_ENV=production` failsafe in the router still throws regardless. Comments in both `docker-compose.yml` and `.env.example` updated to reflect the new opt-in flow.

aspiers:
Urgent bugfix with very heavy testing, merging to get more exposure in test environments as discussed with the team today.

PR #148, opened 2026-05-04
Fix OAuth account enrichment with tooltip explanation

### LINKED ISSUE #143: OAuth enrichment: cover all pages, hide handles in random mode, explain via tooltip
## Bugs

The DOM enrichment script (`buildChooserEnrichmentScript` in `packages/pds-core/src/chooser-enrichment.ts`, mounted by `createChooserEnrichmentMiddleware` and rendered into `/oauth/authorize` + `/account*` HTML responses) has three related defects:

### 1. Wrong scope — only enriches account-chooser rows

The script walks every text node matching a known handle and rewrites the surrounding DOM (inserts `<span class="epds-email-label">…</span>`, tags the matched span with `epds-handle-label`). On the consent screen, on the standalone `/account*` SPA pages, and anywhere else upstream `oauth-provider-ui` shows the user's handle, the same handle text appears in prose copy — `alice.test grants access to ...`, headings, account-management cards, etc. Two failure shapes:

- **Over-enrichment**: handle text in prose got mutated, injecting an email span into copy that was not an account row. This was the original symptom; PR #137 narrowed the rewrite to nodes whose `closest('[role="button"][tabindex="0"]')` resolved (the chooser-row markup oauth-provider-ui 0.4.3 emits).
- **Under-enrichment**: PR #137's narrow scope means the consent screen, `/account*` cards, and any other page where we *do* want the email-first presentation now go un-enriched. We need enrichment on every page that shows the user's handle, not only chooser rows.

### 2. Random-handle mode should hide the handle entirely

`epds_handle_mode` resolves to one of `random`, `picker`, `picker-with-random` (`packages/shared/src/handle.ts`). When the active mode is `random` the handle is an opaque autogenerated string with no user value — surfacing it in the UI is just noise. In that mode the handle span should be hidden (display:none / aria-hidden) and only the email / display name should remain visible.

The other two modes (`picker`, `picker-with-random`) still show the handle as today.

### 3. Tooltip explaining what a handle is

When the handle is hidden (random mode) — and arguably any time we replace handle-prominent UI with email-prominent UI — hovering or focusing the email / account name should reveal a tooltip that:

- Shows the underlying handle value.
- Includes a short, user-friendly explanation of what an AT Protocol handle is and why we have one (e.g. one or two plain sentences plus optionally a link to docs).

Tooltip needs keyboard + screen-reader access (focus trigger, `aria-describedby` or live-region equivalent), not pure `:hover`.

## Scope of the fix

- Generalise the enrichment walker so it covers every page where upstream renders the handle, not just chooser rows. PR #137's row-scoping needs to be reworked into a *positive* allowlist of contexts (chooser row, consent paragraph, account-card, etc.) or a *negative* exclusion list — whichever keeps over-enrichment from coming back.
- Plumb the active `HandleMode` (resolved per-flow already, see `auth_flow.handle_mode` and `resolveHandleMode`) through to the injected script so it can branch on `random` vs picker modes.
- Add the tooltip component (HTML + CSS + minimal JS in the same injected script) with copy that explains handles. Coordinate copy with whoever owns user-facing strings.
- Regression coverage in `packages/pds-core/src/__tests__/chooser-enrichment.test.ts` (and equivalent for `/account*` if not already covered): chooser row, consent prose, multiple account cards, random mode hides handle, picker mode keeps handle, tooltip is reachable by keyboard.

## Related

- PR #137 — partial fix (scopes enrichment to chooser rows). Closes the over-enrichment half but leaves the under-enrichment, random-mode, and tooltip work open.

Kzoeps:
## Summary
This PR restores account identity enrichment across the OAuth and account-management surfaces for ePDS deployments that use generated/random handles.
The user-facing goal is simple: when a handle is system-generated and not meaningful to the user, ePDS should identify the account primarily by email while still keeping the public AT Protocol handle available where it is useful for context, accessibility, or app interoperability.
closes https://github.com/hypercerts-org/ePDS/issues/143

## What changed
- Restores scoped account identity enrichment on:
  - OAuth account chooser pages
  - OAuth consent / approval pages
  - `/account`
  - `/account/:did`
  - preview chooser and preview consent routes

- Makes random-handle flows email-first:
  - chooser rows show email as the visible primary identifier
  - underlying handles are preserved through accessible descriptions
  - consent pages show email as the primary account identifier
  - public handles remain available through the identity information tooltip

- Tightens the DOM enrichment logic so only known account identity elements are rewritten.
  This avoids accidentally mutating legal copy, technical details, connected-app rows, device rows, footer text, or arbitrary prose that happens to contain handle-like text.

- Makes `epds_handle_mode` resolution consistent between auth-service and pds-core:
  - explicit authorize/callback query values take precedence
  - OAuth client metadata is used when the authorize URL only has a PAR `request_uri`
  - valid handle mode is preserved through `/oauth/epds-callback`
  - invalid callback values are ignored instead of forwarded

- Shares OAuth request-context resolution between chooser enrichment and client CSS injection so PAR-backed authorize pages can still resolve the client id correctly.

- Updates random-handle e2e coverage to assert the accessible hidden-handle behavior instead of the previous stale `title` tooltip behavior.

## Summary by CodeRabbit

* **New Features**
  * Account email is now displayed more consistently across sign-in, consent, account chooser, and account management screens.
  * Consent identity information is available through an accessible tooltip, including email and public handle details where relevant.
  * OAuth flows now preserve valid handle-selection modes across approval and account selection.
* **Bug Fixes**
  * Prevented random handles from appearing as primary consent identifiers.
  * Improved recovery when authorization requests provide limited client information.
  * Sign-in actions remain disabled until the page is ready, reducing failed submissions.
* **Accessibility**
  * Hidden handles are now exposed through accessible descriptions instead of title-only tooltips.

## UI Changes
Screenshot_05-May_17-29-13_4449

*when handle mode is picker or picker with random*

Screenshot_05-May_17-31-05_26694
 
*when handle mode is random*


e2e/step-definitions/consent.steps.ts
```diff
@@ -99,6 +129,69 @@ When(
...
+    expect(describedBy?.trim()).toBeTruthy()
+
+    await tooltipControl.click()
+    await expect(tooltipControl).toHaveAttribute('aria-expanded', 'true')
+
+    const tooltip = page.locator(`#${describedBy?.trim()}`)
+    await expect(tooltip).toHaveAttribute('role', 'tooltip')
+    await expect(tooltip).toBeVisible()
+    await expect(tooltip).toContainText('Public AT Protocol handle:')
+    await expect(tooltip).toContainText(publicHandle)
```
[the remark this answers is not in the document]
Kzoeps: _(reply generated by gpt-5.5)_

Fixed in `49c64e1`. Split `aria-describedby` on whitespace and use the first token before constructing the tooltip locator.

Kzoeps:
_(reply generated by gpt-5.5)_

Review follow-up status:

- Fixed in `49c64e1`: `aria-describedby` token handling, chooser-enrichment debug logging, Sonar mechanical code-smell items, test HTTP hotspot annotation, complete-route complexity refactor, session-reuse hidden-handle assertion tightening, and chooser-enrichment test duplication refactor.
- Addressed earlier in `60aa045`: consolidated the generated-handle changeset wording.
- Leaving the `epds_handle_mode` HMAC suggestions as-is by maintainer decision: this field is UI/display-only and is not used as an authorization/security decision.
- Blacksmith E2E failures are intentionally left for separate follow-up per maintainer.

Kzoeps:
_(reply generated by gpt-5.5)_

Follow-up: added `02b4c01` to further reduce the remaining chooser-enrichment test duplication reported by Sonar after the first fix. Local validation for that follow-up passed: `pnpm format:check`, `pnpm lint`, `pnpm typecheck`, and `pnpm vitest run packages/pds-core/src/__tests__/chooser-enrichment.test.ts`.

Kzoeps:
_(reply generated by gpt-5.5)_

Follow-up on the remaining non-thread review surfaces:

- CodeRabbit review https://github.com/hypercerts-org/ePDS/pull/148#pullrequestreview-4226563492 and Sonar comment https://github.com/hypercerts-org/ePDS/pull/148#issuecomment-4377850706: prepared a local fix replacing the two no-interpolation `RegExp` constructors in `e2e/step-definitions/consent.steps.ts` with regex literals. Local `pnpm lint` and `pnpm typecheck` pass. This still needs a commit/push before Sonar can clear it.
- Blacksmith comment https://github.com/hypercerts-org/ePDS/pull/148#issuecomment-4377150944: the failed step timed out waiting for Mailpit before the new consent assertions ran, so I did not change product/test logic for it. The current E2E rerun is still in progress: https://github.com/hypercerts-org/ePDS/actions/runs/25367208712.
- No unresolved inline review threads remain.

Kzoeps:
@coderabbitai review

packages/auth-service/src/routes/complete.ts
```diff
@@ -166,6 +180,7 @@ export function createCompleteRouter(
           ctx.config.epdsCallbackSecret,
         )
         const params = new URLSearchParams({ ...callbackParams, ts, sig })
+        params.set('epds_handle_mode', flow.handleMode)
```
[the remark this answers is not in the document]
Kzoeps: does it make sense to add it to  callback params when its just for cosmetic purposes and doesnt really tie into any security issues. Or is it better for consistency to just add it in the signCallback params

Kzoeps:
clsoing and reopening for ci to run properly

packages/auth-service/src/routes/complete.ts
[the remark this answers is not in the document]
Kzoeps: added it to the HMAC signed callback params to maintain consistency

Kzoeps:
@coderabbitai please review

Kzoeps:
_(reply generated by gpt-5.5)_

Review follow-up status for the current PR comment pass:

- CodeRabbit review https://github.com/hypercerts-org/ePDS/pull/148#pullrequestreview-4228468365: local fixes are prepared for the still-valid items. `choose-handle.ts` now signs `epds_handle_mode` as an empty sentinel when absent, and `consent.steps.ts` now uses one `openIdentityTooltip()` helper. The `complete.ts` callback-signing item was already addressed in `7396050`; the `client-css-injection.ts` logger item was already addressed in `e0b618f`.
- Earlier CodeRabbit summaries https://github.com/hypercerts-org/ePDS/pull/148#pullrequestreview-4225652757 and https://github.com/hypercerts-org/ePDS/pull/148#pullrequestreview-4226563492 are stale/already addressed by existing commits plus the local duplication refactors.
- Sonar comment https://github.com/hypercerts-org/ePDS/pull/148#issuecomment-4380772977: local duplication-reduction changes are prepared in `crypto.test.ts`, `consent.steps.ts`, and `callback-handle-mode.test.ts`; Sonar cannot re-evaluate until these are committed and pushed.
- Blacksmith failure https://github.com/hypercerts-org/ePDS/pull/148#issuecomment-4380322626 appears stale; latest `gh pr checks` after `265df25` shows all visible checks except Sonar passing.

Local validation after the prepared changes passed: `pnpm format:check`, `pnpm lint`, `pnpm typecheck`, and `pnpm vitest run packages/shared/src/__tests__/crypto.test.ts packages/auth-service/src/__tests__/callback-handle-mode.test.ts`.

Blocked: the code changes are intentionally uncommitted/unpushed pending human approval.

Kzoeps:
@coderabbitai please review

Kzoeps:
@coderabbitai please review

packages/auth-service/src/routes/login-page.ts
```diff
@@ -1162,6 +1164,9 @@ export function renderLoginPage(opts: {
         clearOtpBoxes();
       });
 
+      sendOtpForm.dataset.epdsLoginReady = 'true';
+      sendOtpBtn.disabled = false;
```
aspiers: @Kzoeps These seem to trigger unconditionally on page load, which would make this disable/enable mechanism ineffective.

.changeset/clean-exit-on-expired-signin.md
aspiers, line 1: I don't know why this PR is removing these files, as they don't even exist on `main` any more.  Perhaps it's a bug in the way GitHub is displaying it, and a rebase on latest `main` will make them go away.

packages/pds-core/src/lib/epds-callback-authorize.ts
```diff
@@ -0,0 +1,26 @@
...
+export function resolveCallbackHandleMode(
+  value: unknown,
+): HandleMode | undefined {
+  return typeof value === 'string' &&
+    (VALID_HANDLE_MODES as readonly string[]).includes(value)
+    ? (value as HandleMode)
+    : undefined
+}
+
+export function buildEpdsCallbackAuthorizeUrl(opts: {
```
aspiers: This seems a little odd.  Surely the code already has had to construct callback URLs for a long time, so why is a new function needed now?

CHANGELOG.md
aspiers, line 1: Yeah this definitely shouldn't be here, it's already in `main` 🤷🏼

Kzoeps: resolved after updating branch

packages/auth-service/src/routes/login-page.ts
aspiers: Agreed, and this was addressed later in the branch (`efa6e92`, then `802a46b`): the `data-epds-login-ready` / `epdsLoginReady` marker you were looking at is gone entirely.

What is left is just the one line: the button is rendered server-side as `<button type="submit" class="btn-primary" disabled>Continue</button>` and the inline script sets `sendOtpBtn.disabled = false;` only after the last `addEventListener` call, at top-level script scope. So it is not triggered on page load independently of handler installation — it *is* the end of handler installation. A `<noscript>` warning covers the JS-disabled case.

`login-page.test.ts` pins this from both directions: it asserts the marker is absent, that the button renders disabled, and that the enable appears after the last click handler.

One related change I had to make while rebasing: the accompanying `does not use a blind timer for readiness` test banned `setTimeout`/`setInterval` across the whole rendered page, which now trips on the unrelated PAR-heartbeat `setInterval` that landed on main. I scoped it to the gap between the last handler registration and the enable, which is the actual invariant being defended.

packages/pds-core/src/lib/epds-callback-authorize.ts
aspiers: Fair challenge. You are right that the callback URL construction itself is long-standing and trivial — before this PR it was three inline lines in `pds-core/src/index.ts`:

```ts
const authorizeUrl = new URL('/oauth/authorize', pdsUrl)
authorizeUrl.searchParams.set('request_uri', requestUri)
authorizeUrl.searchParams.set('client_id', clientId)
```

A function purely for that would be gratuitous. What justifies extracting it is the thing this PR adds: `epds_handle_mode` has to be forwarded onto that redirect *and validated*, because the mode arrives from a query parameter and must not be forwarded when it is not a recognised mode. That is real branching logic worth unit-testing without standing up the whole callback route.

It was also badly named, which is probably what made it look redundant — `buildEpdsCallbackAuthorizeUrl` read as "another epds-callback URL builder" alongside the auth-service ones. It was renamed to `buildPostCallbackAuthorizeUrl` in `802a46b` with a docstring stating the direction explicitly: this builds the *outbound* redirect from `/oauth/epds-callback` back to stock `/oauth/authorize`, unlike the auth-service builders which construct signed inbound auth-service → pds-core URLs.

`epds-callback-authorize.test.ts` covers both directions (valid mode preserved, invalid mode dropped). I also checked the empty-string case, since `choose-handle` signs `epds_handle_mode: flow.handleMode ?? ''`: `''`, `undefined` and `'garbage'` all resolve to no parameter, while `random` / `picker` / `picker-with-random` pass through. So the sentinel round-trips through signing without leaking a bogus mode downstream.

If you would still rather see it inlined now that the naming is clearer, say so and I will fold it back.

aspiers: Separate note while rebasing, on the neighbouring (already-resolved) CodeRabbit thread about logging the request-uri/metadata fallback: the `logger?.debug` call and its unit test were both in place, but the production call site in `pds-core/src/index.ts` never passed a `logger`, so the optional chain silently made it a no-op in every real deployment. Fixed in `1104efa` by threading the existing pds-core logger through. Flagging it here since that thread was marked resolved while the logging was still dead in production.

aspiers:
### Note on the e2e failure seen on this branch (resolved — flaky, not a regression)

The first E2E run on `1104efa` failed one scenario of 72, `Expired OTP is rejected, resend recovers the flow`:

```
Locator: locator('#error-msg')
Expected substring: "OTP expired"
Received string:    "Invalid OTP"
```

Re-running the same job on the identical commit passed, so this is flake rather than a regression from the rebase. All checks on `1104efa` are now green. Recording the diagnosis since the symptom is misleading and will recur.

**Why it looks like a product bug but isn't:** better-auth throws `Invalid OTP` both for a wrong code *and* for no pending code at all — noted already in `packages/auth-service/src/better-auth.ts` above `OTP_FAILURE_MESSAGES`. An absent verification row and a mistyped code are therefore indistinguishable in `#error-msg`.

**Suspected root cause, unfixed and pre-existing:** `POST /_internal/test/expire-otp` in `packages/auth-service/src/routes/test-hooks.ts` runs

```sql
UPDATE verification SET expiresAt = ? WHERE identifier = ?
```

with no liveness guard, whereas the sibling `expire-auth-flow` hook a few lines above *does* filter `AND expires_at > ?`. The step asserts only `updated >= 1`, which a stale row can satisfy while the live verification path finds no pending OTP — yielding `Invalid OTP` instead of `OTP expired`.

I have deliberately not fixed this here: it is unrelated to this PR's scope and would widen the diff. Worth a separate issue if the scenario keeps failing.


Tracked separately as #214 (with the better-auth `INVALID_OTP` vs `OTP_EXPIRED` branch logic and a suggested one-line hook fix). Not addressed in this PR.

packages/pds-core/src/chooser-enrichment.ts
```diff
@@ -213,19 +548,25 @@ export function buildChooserEnrichmentScript(): string {
...
+          var descriptionId = 'epds-hidden-handle-' + matches.indexOf(m);
+          description.id = descriptionId;
+          description.className = 'epds-hidden-handle-description';
+          description.textContent = 'Underlying handle: ' + ownText;
+          description.style.cssText =
+            'position:absolute;width:1px;height:1px;padding:0;margin:-1px;overflow:hidden;clip:rect(0,0,0,0);white-space:nowrap;border:0;';
+          description.setAttribute('aria-hidden', 'false');
+          row.appendChild(description);
+          appendAriaReference(row, 'aria-describedby', descriptionId);
         }
```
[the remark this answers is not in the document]
aspiers: _(reply generated by Claude Opus 5 via Claude Code)_

Fixed in 6b15462. Confirmed the mechanism before changing anything: `enrich()` rebuilds `matches` on every `tick()`, `tick()` is driven by a `MutationObserver` on the whole document subtree, and already-enriched rows are excluded via `dataset.epdsEnriched` — so a row enriched on a later re-render does restart at index 0 and collide. Real a11y bug, since duplicate ids make `aria-describedby` resolve to the first node and a row announces another account's handle.

Used a monotonic `hiddenHandleSeq` declared outside `enrich()` so it survives across ticks (inside the function it would reset every tick and change nothing).

Added a regression test that replays an observer tick: the harness `MutationObserver` stub now optionally hands its callback back to the test, so a second row can arrive after the first is already enriched. Verified it actually catches the bug — against the old `matches.indexOf(m)` it fails with `expected 'epds-hidden-handle-0' not to be 'epds-hidden-handle-0'`, and passes with the counter.

One incidental gotcha worth recording: this file is a `String.raw` template producing browser JS, so backticks in comments become interpolation delimiters. My first draft of the explanatory comment used `` `matches` `` and broke the build with `Expected ";" but found "matches"`.

packages/auth-service/src/__tests__/login-page.test.ts
```diff
@@ -521,6 +521,47 @@ function renderDefault(): string {
...
+    // the enable is driven by handler setup completing, not by a timer.
+    const enableIdx = html.indexOf('sendOtpBtn.disabled = false;')
+    const lastHandlerIdx = html.lastIndexOf("addEventListener('click'")
+
+    expect(lastHandlerIdx).toBeGreaterThan(0)
+    expect(enableIdx).toBeGreaterThan(lastHandlerIdx)
+    expect(html.slice(lastHandlerIdx, enableIdx)).not.toMatch(
+      /set(?:Timeout|Interval)\s*\(/,
+    )
+  })
```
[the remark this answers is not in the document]
aspiers: _(reply generated by Claude Opus 5 via Claude Code)_

Fixed in 6b15462 — you are right, and this was my own change during the rebase, so the criticism lands squarely.

Verified the ordering in the generated page: the submit binding is at `sendOtpForm.addEventListener('submit'` (~line 1134), the last click handler is `btn-back` (~1264), and the enable is at ~1271. So anchoring on the last *click* handler did not prove anything about the submit binding — moving it below the enable would still have passed.

Both tests now anchor on the submit binding. I kept the last-click-handler assertion in `enables the button after handler setup` as a *second* check rather than replacing it, since the enable being last in handler setup is also worth pinning; the timer test scopes its `setTimeout`/`setInterval` scan to the submit-binding→enable window. That window still excludes the PAR-heartbeat `setInterval` (~line 804), which was the reason the original page-wide ban had to be narrowed in the first place.

e2e/step-definitions/consent.steps.ts
```diff
@@ -35,6 +36,59 @@ import { fillOtp } from '../support/otp.js'
...
+  const [tooltipId] = describedBy.trim().split(/\s+/)
+
+  await tooltipControl.click()
+  await expect(tooltipControl).toHaveAttribute('aria-expanded', 'true')
+
+  const tooltip = page.locator(`#${tooltipId}`)
+  await expect(tooltip).toHaveAttribute('role', 'tooltip')
+  await expect(tooltip).toBeVisible()
+  return tooltip
+}
```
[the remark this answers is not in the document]
aspiers: _(reply generated by Claude Opus 5 via Claude Code)_

Fixed in 6b15462 with `.first()`.

Checked the risk is real rather than theoretical: `enrichConsentIdentity()` walks every node via `createTreeWalker` and calls `appendIdentityInfoIcon` on each match with no early exit, and `hasApprovedConsentIdentityContext()` approves both `"Grant access to your ... account"` and `"... wants to access your ... account"`. So a page rendering both phrasings genuinely gets two icons. The unit-test helper `expectConsentTooltipTexts()` already takes an *array* of expected texts, which confirms multiples are an expected state — the e2e helper was the one assuming a single match.

Added a comment explaining why `.first()` is well-defined here: each icon carries its own `aria-describedby`, so the helper still asserts against a coherent control/tooltip pair rather than papering over an ambiguous locator.

packages/pds-core/src/lib/oauth-request-context.ts
```diff
@@ -0,0 +1,37 @@
...
+): Promise<string | undefined> {
+  if (typeof query.client_id === 'string') return query.client_id
+  if (!resolveClientIdFromRequestUri) return undefined
+
+  const requestUri =
+    typeof query.request_uri === 'string' ? query.request_uri : undefined
+  if (!requestUri) return undefined
+
+  return resolveClientIdFromRequestUri(requestUri)
+}
```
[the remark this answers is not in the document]
aspiers: _(reply generated by Claude Opus 5 via Claude Code)_

Added in 6b15462 — new `packages/pds-core/src/__tests__/oauth-request-context.test.ts`, 9 cases covering both paths you named plus the edges:

- explicit `client_id` returned directly, and preferred over a present `request_uri` (asserting the resolver is *not* called, so the short-circuit is pinned rather than just the return value)
- `request_uri` resolved through the supplied resolver, asserting it receives the right argument
- resolver returning `undefined`
- no resolver supplied
- empty query
- non-string `client_id` / `request_uri` ignored (the `typeof` guards)
- resolver rejection propagating, which pins the documented contract that error handling is deliberately left to each middleware

packages/auth-service/src/routes/login-page.ts
aspiers: _(reply generated by Claude Opus 5 via Claude Code)_

Leaving this thread open for @aspiers to confirm, since the reply above is my assessment that your original concern is now satisfied rather than a change you asked for — you should be the one to decide whether it actually is.

Related update: CodeRabbit made a fair point on the accompanying test, that anchoring on the last click handler did not prove the *submit* binding precedes the enable. Fixed in 6b15462 so both readiness tests assert against `sendOtpForm.addEventListener('submit'` directly.

packages/pds-core/src/lib/epds-callback-authorize.ts
aspiers: _(reply generated by Claude Opus 5 via Claude Code)_

Leaving this thread open deliberately: my earlier reply argues the extraction is justified and offers to inline it again, which is a disagreement with your original point rather than a fix. That is your call to make, so self-resolving it would hide an open question. Resolve it if the rename plus docstring answers the objection; say the word and I will fold the function back inline.

aspiers:

### Re: the four nitpicks in CodeRabbit's latest review summary

These were in the review body rather than as inline threads, so replying here. Three addressed, one declined.

**1. Escape-to-dismiss for the identity tooltip** (`chooser-enrichment.ts`) — **fixed in 7f76cd6.** Valid and worth more than "trivial": the tooltip had no key handler at all, and `hide()` returns early while `pinned`, so a keyboard/AT user who clicked to pin it genuinely could not dismiss it without moving focus back to the button. That is the WCAG 1.4.13 Dismissible failure as described. Escape now clears `pinned` then hides, bound on `keyup` so it does not race the surrounding page for the same key, and it no-ops when the tooltip is already hidden rather than swallowing the event. Covered by a new test that asserts the pinned path, the unpinned path, a non-Escape key, and non-propagation when already hidden — verified it fails against the pre-fix code.

**2. Hoist the duplicated handle-mode resolver** (`preview-consent.ts`) — **fixed in ea43ebc.** This PR added a byte-identical second copy of `resolveQueryHandleMode` next to the existing one in `preview-chooser.ts`. Since both previews exist specifically to exercise the production resolver path, two copies is a drift risk rather than harmless repetition. Moved into `preview-shared.ts`, which already owns the helpers both routes share, and dropped the now-dead `resolveHandleMode` / `VALID_HANDLE_MODES` / `ClientMetadata` imports from both call sites.

**3. Document the fake `createTreeWalker` divergence** (`chooser-enrichment.test.ts`) — **done in 7f76cd6.** Agreed with your own assessment that it is benign today: nothing the script inserts can match `isConsentIdentityElement()` (which needs a `<b>`/`<strong>` inside an approved consent phrasing) or `matchAccountIdentifier()`. Added a comment recording that the snapshot-vs-live difference is deliberate and that the fake should be made live if either predicate loosens, so the limitation is not silently trusted later.

**4. Use a CSS class for runtime visibility** (`login-page.ts:879-917`) — **declined, out of scope.** `refreshResendVisibility()` is pre-existing code from main, not part of this PR: `git diff origin/main HEAD -- packages/auth-service/src/routes/login-page.ts` adds zero lines touching `style.display`. Inline `style.display` is also the established idiom throughout that file (7 call sites, including `errorEl` and `termsEl`), so converting only the three lines you flagged would leave the file half-migrated and inconsistent. Worth doing as its own change across the whole file if the guideline is to be enforced, but folding it into an OAuth-enrichment PR would mix unrelated churn into the diff.

All green on 7f76cd6 locally: build, lint, format, and 1132 unit tests (Node 20, matching CI).

aspiers:
@Kzoeps I've split this into four stacked PRs to make it reviewable.

The motivation: this PR is +2918/−173, but ~1450 of that is a single test file, so the real change is ~700 lines covering four independent concerns. In particular the HMAC payload change is currently reviewed alongside 1400 lines of DOM tests, and it deserves its own scrutiny.

| PR | Change | Base |
|---|---|---|
| #241 | Extract OAuth client-id resolution from the CSS middleware | `main` |
| #242 | Gate email sign-in until its handlers are installed | `main` |
| #243 | Sign `epds_handle_mode` through the callback hop | #241 |
| #244 | Identify generated-handle accounts by email | #243 |

#242 is entirely independent and could merge today. #241 is a near-trivial refactor. #243 carries the HMAC change. #244 is the actual user-visible enrichment work.

**Nothing was changed in the process.** Merging all four branches produces a tree byte-identical to this PR's head (`7fe24bb`) — verified by diffing the union against it. The only difference is the single `sign-in-account-presentation.md` changeset, which I split into three so each PR carries the part describing its own change (#241 needs none — it's an internal refactor).

Each branch independently passes `typecheck`, `lint`, `format` and the full vitest suite.

Two things I flagged in the split that are worth your eyes:

- **#243 changes the signed HMAC payload**, so auth-service and pds-core must be deployed together. In-flight callbacks fail verification across a staggered rollout. Detail in that PR.
- **#243 also carries the `resolveCompleteIdentity()` extraction** from `/auth/complete`, which is unrelated to handle mode. I kept it rather than dropping work you'd done, but it's easy to pull out if you'd prefer.

Your two review threads here are both marked outdated; the `login-page.ts` one ("these seem to trigger unconditionally on page load") looks answered by `69a1456`, which is the disabled-button gating now in #242.

PR #165, opened 2026-05-07
[WIP] many usability fixes (needs untangling)

aspiers:
**WIP CAVEAT: This was originally a few simple fixes which have since landed separately in #187.  However an over-zealous AI agent was given a bit too much freedom and found a ton of other things to add, some of which are genuine improvements, and others of questionable benefit.**

So we need to triage the remaining commits and decide which to keep and which to ditch.  They should certainly be split up into smaller PRs.

**N.B. The rest of this description is probably out of date!**

------

## Summary

- Hide the **Resend code** button on the OTP screen the moment the sign-in can no longer recover; surface **Start over** in its place. The screen never offers an action that cannot complete the flow.
- Inline action on the OTP-expired error now branches the same way: **Start over** when the flow is dead, **Send a new code** when it can still finish.
- Reconciles on heartbeat ticks (including transient ones) and on the `visibilitychange` event, so a backgrounded tab returning to focus reflects reality immediately.

## Why

Bug report: user requested an OTP, waited ~11 minutes, got "OTP expired" (correct), clicked **Send a new code**, received a fresh email, typed it, and only then saw "Sign in failed". A fresh code was issued that could never have worked because the upstream PAR row had silently lapsed during the wait.

The earlier reactive abort gate (PR #154) bounces the user cleanly to the OAuth client when this happens, but only after the user has already invested time typing a code that was never going to succeed. This change makes the offer disappear before the user can act on it, so the only forward paths surfaced are paths that actually work.

## How

- New `lastSuccessfulHeartbeatAt` timestamp updated on every `ok:true` ping. Initialised to page-load time (PAR is fresh on render).
- New `parLikelyDead()` returns true when `now - lastSuccessfulHeartbeatAt >= 5min` (upstream's `AUTHORIZATION_INACTIVITY_TIMEOUT`) or when the page has already been told the flow aborted.
- New `refreshResendVisibility()` toggles `#btn-resend` ↔ `#btn-start-over` based on `parLikelyDead()`. Called from heartbeat tick, visibilitychange, OTP-step show, and the inline-action error path.

## Out of scope

Server-side enforcement of the same invariant — `/email-otp/send-verification-otp` and `/sign-in/email-otp` will still process direct POSTs that bypass the UI. Worth doing as a defence-in-depth follow-up; not load-bearing for honest-user UX, which is what this PR addresses.

## Test plan

- [x] New `@resend-hidden-when-par-dead` scenario passes.
- [x] Full `@otp-and-par-expiry` / `@par-heartbeat` / `@resend-after-par-dead` / `@otp-expiry` suite passes (7 scenarios, 78 steps).
- [x] `pnpm format:check`, `pnpm lint`, `pnpm typecheck`, `pnpm test` (1017 unit tests) all green locally.
- [ ] CI on this branch.

## Summary by CodeRabbit

* **New Features**
  * OTP UI shows “Start over” when a sign-in can’t be recovered; inline “Send a new code” on lockout. Demo now surfaces “session expired”; OAuth session cookie extended to 1 hour. Account Settings shows visible success/error flash banners.

* **Bug Fixes**
  * Don’t flash “Invalid OTP” on empty submit; focus first OTP box. Clear OTP boxes on resend; paste/keystroke filtered to OTP charset; SMS autofill distributes across boxes. Recovery links reliably return to the originating sign-in; friendlier error and lockout wording. Reserved handles marked unavailable.

* **Tests**
  * Expanded E2E and unit tests for OTP, PAR liveness, recovery, and error flows.


aspiers:
Updating description regarding the status of this PR, and converting to draft.

aspiers:
Superseded by many smaller PRs.

PR #188, opened 2026-07-07
Replace transition:generic with permission sets scopes

Ashex:
Drop legacy scopes and use permission sets introduced in https://github.com/hypercerts-org/hypercerts-lexicon/pull/222

### DIFF
--- a/.agents/skills/epds-login/SKILL.md
+++ b/.agents/skills/epds-login/SKILL.md
@@ -15,6 +15,11 @@ AT Protocol universe (a DID, a handle, a data repository) automatically provisio
 From your app's perspective, ePDS uses standard AT Protocol OAuth (PAR + PKCE + DPoP).
 The reference implementation is `packages/demo` in the [ePDS repository](https://github.com/hypercerts-org/ePDS).
 
+For protocol-level detail beyond ePDS specifics — DPoP proof mechanics, granular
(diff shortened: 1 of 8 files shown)


aspiers:
I tested the untrusted demo service and this looks good:

image

aspiers:
One thing worth adding while we're looking at this dialog: since the point of the demo is to show developers the consent screen a *typical* Hypercerts app produces, it should request the scopes such an app actually needs — otherwise the dialog under-represents what real users will see.

The two `include:*.authWrite` sets bundle only record writes (create/update/delete), so a capability most HC apps need is missing:

**Blob upload — `blob:*/*`.** None of the permission sets grant blob upload, but most HC apps upload media (logos, images, attachments → `org.hypercerts.context.attachment` etc.), and a record write referencing a blob fails without an upload grant. Adding `blob:*/*` (any MIME type) restores the capability *and* surfaces the upload consent card in this dialog.

Optionally, `account:email` (read is the default action) would add an email card too — grants email + verification status via `getSession` — if a typical HC app would want the user's email for notifications/account linking. No legacy `transition:email` needed.

Suggested scope:
```
atproto include:org.hypercerts.authWrite include:app.certified.authWrite blob:*/*
```
(add `account:email` if wanted). The consent e2e (`consent.steps.ts`) could then also assert the blob card renders.

aspiers (approved):
Added some suggestions for improvements but this is already a good step in the right direction 👍🏼

PR #204, opened 2026-07-20
auth: support mobile paste in segmented code input

### LINKED ISSUE #212: Mobile OTP paste and accessibility
The OTP UI uses a separate `maxlength="1"` input for each digit. On mobile—especially iOS—pasting or using one-time-code autofill can place the full code into only the focused field, causing it to be truncated instead of filling every slot.

The segmented inputs also expose one verification code as several unrelated fields to assistive technology. Screen-reader navigation is noisy, automatic submission is not explained, and status or error updates are not reliably announced. The component should keep its segmented appearance while behaving as one accessible input with reliable mobile paste and autofill.

Kzoeps:
## Summary

* replace per-character OTP inputs with one full-length input behind the existing visual slots
  * this is how its done in shadcn and is the recommended way for accessbility. Heavily referenced from [input-otp](<https://github.com/guilhermerodz/input-otp>)
* preserve mobile paste, one-time-code autofill, and keyboard replacement behavior
* keep client branding selectors and browser-driven OTP helpers compatible
* add pure normalization and selection tests plus a patch changeset

Closes hypercerts-org/ePDS#212

## Testing

* `pnpm typecheck`
* `pnpm lint`
* `pnpm test` — 1,057 tests passed
* `pnpm test:coverage`
* Chromium mobile-context smoke test for formatted paste and keyboard replacement

## Summary by CodeRabbit

* **Bug Fixes**
  * Pasting or autofilling a mobile sign-in code now fills the complete one-time code instead of only the first character.
  * Improved OTP input behavior, including caret movement, selection, validation, normalization, and automatic submission.
  * OTP controls now receive clearer focus, accessibility, and disabled-state handling.
* **Style**
  * Updated OTP active-state styling guidance for themed interfaces.

## Summary by CodeRabbit

* **New Features**
  * Improved one-time-code entry on mobile, including full-code paste and autofill.
  * Added easier selection and replacement of individual code positions.
  * Improved screen-reader announcements and support for reduced-motion preferences.
  * OTP entry now uses a single input while preserving the segmented visual design.
* **Bug Fixes**
  * Corrected code normalization, caret movement, focus handling, and automatic submission.
  * Updated active-state styling for OTP fields.

## Summary by CodeRabbit

* **New Features**
  * Improved mobile sign-in code entry with long-press paste and one-time-code autofill.
  * Added flexible code selection, replacement, normalization, and automatic submission.
  * Added screen-reader announcements, accessibility metadata, and reduced-motion support.
* **Style**
  * Improved compatibility with existing sign-in branding and input styles.
  * Added clearer focus, placeholder, active-slot, and caret styling.
* **Documentation**
  * Updated guidance for OTP styling and browser-based code entry.

### DIFF
--- a/.changeset/mobile-code-paste.md
+++ b/.changeset/mobile-code-paste.md
@@ -6,6 +6,6 @@ Pasting a sign-in code on mobile now fills the complete code instead of only the
 
 **Affects:** End users, Client app developers
 
-**End users:** No action is required; long-press paste and mobile one-time-code autofill now work with the existing segmented code-field appearance. Tapping or clicking a filled code slot selects that character for replacement.
+**End users:** No action is required; long-press paste and mobile one-time-code autofill now work with the existing segmented code-field appearance. Tapping or clicking a filled code slot selects that character for replacement. The field now provides clearer keyboard focus, announces status messages to screen readers, describes automatic submission, and respects reduced-motion preferences.
 
 **Client app developers:** Branding CSS that styles an active code slot should use `.otp-box.active` instead of `.otp-box:focus`; the real input now keeps focus while the slots remain presentational.
--- a/.changeset/mobile-code-paste.md
+++ b/.changeset/mobile-code-paste.md
@@ -1,11 +1,15 @@
 ---
-'ePDS': patch
+'ePDS': minor
 ---
 
-Pasting a sign-in code on mobile now fills the complete code instead of only the first character.
+Entering a sign-in code now works better on phones and with assistive technology.
 
 **Affects:** End users, Client app developers
 
-**End users:** No action is required; long-press paste and mobile one-time-code autofill now work with the existing segmented code-field appearance. Tapping or clicking a filled code slot selects that character for replacement. The field now provides clearer keyboard focus, announces status messages to screen readers, describes automatic submission, and respects reduced-motion preferences.
+**End users:** Long-press paste and one-time-code autofill now fill the whole code. Filled positions can be selected and replaced, status messages are announced to screen readers, and the code field respects reduced-motion settings.
 
-**Client app developers:** Branding CSS that styles an active code slot should use `.otp-box.active` instead of `.otp-box:focus`; the real input now keeps focus while the slots remain presentational.
+**Client app developers:** This is a breaking change for branding CSS that targets the sign-in code field.
+
+- `.otp-box` is now a `<div>` instead of an `<input>`. Selectors such as `input.otp-box`, `input[data-slot]`, and `.otp-box::placeholder` no longer match.
+- `.otp-box:focus` no longer matches because one `.otp-input-overlay` now owns focus. Use `.otp-box.active` for the selected position and `.otp-character.placeholder` for empty positions.
+- Broad `input` and `input:focus` rules now also match `.otp-input-overlay`. Exclude that class or override it so its background, border, outline, text, and caret stay transparent; otherwise it may cover the visible code boxes.
--- a/.changeset/mobile-code-paste.md
+++ b/.changeset/mobile-code-paste.md
@@ -6,6 +6,6 @@ Pasting a sign-in code on mobile now fills the complete code instead of only the
 
 **Affects:** End users, Client app developers
 
-**End users:** No action is required; long-press paste and mobile one-time-code autofill now work with the existing segmented code-field appearance.
+**End users:** No action is required; long-press paste and mobile one-time-code autofill now work with the existing segmented code-field appearance. Tapping or clicking a filled code slot selects that character for replacement.
 
 **Client app developers:** Branding CSS that styles an active code slot should use `.otp-box.active` instead of `.otp-box:focus`; the real input now keeps focus while the slots remain presentational.
--- a/.changeset/mobile-code-paste.md
+++ b/.changeset/mobile-code-paste.md
@@ -8,8 +8,4 @@ Entering a sign-in code now works better on phones and with assistive technology
 
 **End users:** Long-press paste and one-time-code autofill now fill the whole code. Filled positions can be selected and replaced, status messages are announced to screen readers, and the code field respects reduced-motion settings.
 
-**Client app developers:** This is a breaking change for branding CSS that targets the sign-in code field.
-
-- `.otp-box` is now a `<div>` instead of an `<input>`. Selectors such as `input.otp-box`, `input[data-slot]`, and `.otp-box::placeholder` no longer match.
-- `.otp-box:focus` no longer matches because one `.otp-input-overlay` now owns focus. Use `.otp-box.active` for the selected position and `.otp-character.placeholder` for empty positions.
-- Broad `input` and `input:focus` rules now also match `.otp-input-overlay`. Exclude that class or override it so its background, border, outline, text, and caret stay transparent; otherwise it may cover the visible code boxes.
+**Client app developers:** Existing trusted-client visual styling on direct `input` selectors and their focus or placeholder states continues to style the sign-in code slots automatically. Common `input.otp-box`, `input[data-slot]`, `.otp-box:focus`, and `.otp-box::placeholder` forms are also supported. New branding should use `.otp-box.active` for the selected position and `.otp-character.placeholder` for empty positions; `.otp-input-overlay` is internal and protected from broad `input` rules. Browser automation that previously filled each `.otp-box` should fill the single `#code` input instead; malformed CSS, mechanical input declarations, and more complex DOM-coupled selectors remain unchanged.
--- a/.changeset/mobile-code-paste.md
+++ b/.changeset/mobile-code-paste.md
@@ -1,0 +1,11 @@
+---
+'ePDS': patch
+---
+
+Pasting a sign-in code on mobile now fills the complete code instead of only the first character.
+
+**Affects:** End users, Client app developers
+
+**End users:** No action is required; long-press paste and mobile one-time-code autofill now work with the existing segmented code-field appearance.
+
+**Client app developers:** Branding CSS that styles an active code slot should use `.otp-box.active` instead of `.otp-box:focus`; the real input now keeps focus while the slots remain presentational.
--- a/packages/auth-service/src/routes/login-page.ts
+++ b/packages/auth-service/src/routes/login-page.ts
@@ -45,6 +45,7 @@ import { socialProviders } from '../better-auth.js'
 import {
   buildOtpInputProps,
   normalizeOtpValue,
+  resolveOtpClickSelection,
   resolveOtpSelection,
 } from '../otp-input.js'
 import {
@@ -522,6 +523,7 @@ export function renderLoginPage(opts: {
   // It is self-contained, so serializing its source avoids maintaining a
   // second client-only implementation inside this server-rendered page.
   const normalizeOtpValueSource = normalizeOtpValue.toString()
+  const resolveOtpClickSelectionSource = resolveOtpClickSelection.toString()
   const resolveOtpSelectionSource = resolveOtpSelection.toString()
 
   // ATProto/Bluesky handle login button.
@@ -801,6 +803,7 @@ export function renderLoginPage(opts: {
       var otpLength = ${opts.otpLength};
       var otpCharset = ${JSON.stringify(opts.otpCharset)};
       var normalizeOtpValue = (${normalizeOtpValueSource});
+      var resolveOtpClickSelection = (${resolveOtpClickSelectionSource});
       var resolveOtpSelection = (${resolveOtpSelectionSource});
 
       // PAR heartbeat — slides the upstream request_uri inactivity
@@ -1114,7 +1117,29 @@ export function renderLoginPage(opts: {
 
       otpInput.addEventListener('focus', positionOtpSelection);
       otpInput.addEventListener('blur', renderOtpSlots);
-      otpInput.addEventListener('click', syncOtpSelection);
+      otpInput.addEventListener('click', function(e) {
+        // The real input must remain above the visual slots for iOS long-press
+        // paste. Read the box underneath the click without changing hit testing.
+        var clickedBox = document.elementsFromPoint(e.clientX, e.clientY).find(function(element) {
+          return element.hasAttribute('data-slot');
+        });
+        if (!clickedBox) {
+          syncOtpSelection();
+          return;
+        }
+
+        var selection = resolveOtpClickSelection(
+          otpInput.value.length,
+          Number(clickedBox.getAttribute('data-slot')),
+        );
+        otpInput.setSelectionRange(
+          selection.start,
+          selection.end,
+          selection.direction,
+        );
+        previousOtpSelection = [selection.start, selection.end];
+        renderOtpSlots();
+      });
       otpInput.addEventListener('keyup', syncOtpSelection);
       document.addEventListener('selectionchange', function() {
         if (document.activeElement === otpInput) syncOtpSelection();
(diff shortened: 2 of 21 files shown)


Kzoeps:
_(reply generated by OpenAI Codex)_

Addressed the SonarCloud finding reported in [this comment](https://github.com/hypercerts-org/ePDS/pull/204#issuecomment-5021913813) in `af18b8e` — numeric code normalization now uses the concise `\D` character class.

Kzoeps:
@coderabbitai please review

Kzoeps:
@coderabbitai please review

.changeset/mobile-code-paste.md
```diff
@@ -0,0 +1,11 @@
...
+'ePDS': patch
+---
+
+Pasting a sign-in code on mobile now fills the complete code instead of only the first character.
+
+**Affects:** End users, Client app developers
+
+**End users:** No action is required; long-press paste and mobile one-time-code autofill now work with the existing segmented code-field appearance. Tapping or clicking a filled code slot selects that character for replacement. The field now provides clearer keyboard focus, announces status messages to screen readers, describes automatic submission, and respects reduced-motion preferences.
+
+**Client app developers:** Branding CSS that styles an active code slot should use `.otp-box.active` instead of `.otp-box:focus`; the real input now keeps focus while the slots remain presentational.
```
aspiers: @Kzoeps Is this a breaking change which all branded frontends will have to adapt to? If so I think we need to ensure it's backwards compatible, otherwise technically this is a major version change not patch level.  Either way I think this needs wording more clearly.

aspiers (changes requested):
I've tested this on mobile and it works for me, but it needs more clarity around whether it's a breaking change (it shouldn't be).

Also seems like a ton of new code for what I would have expected to be a pretty small tweak, but it seems that's because we're now taking over all control of handling input which was previously handled natively. I would have preferred to see it fixed whilst retaining native input control somehow, but maybe that's just not possible.

packages/auth-service/src/routes/login-page.ts
```diff
@@ -509,6 +514,12 @@ export function renderLoginPage(opts: {
     : `<img src="/static/certified-brandmark.svg" alt="Certified" class="client-logo">`
 
   const inputProps = buildOtpInputProps(opts.otpLength, opts.otpCharset)
+  // The browser uses the exact pure normalizer covered by otp-input.test.ts.
+  // It is self-contained, so serializing its source avoids maintaining a
+  // second client-only implementation inside this server-rendered page.
+  const normalizeOtpValueSource = normalizeOtpValue.toString()
+  const resolveOtpClickSelectionSource = resolveOtpClickSelection.toString()
+  const resolveOtpSelectionSource = resolveOtpSelection.toString()
```
aspiers, line 554: Non-blocking, but worth a guard.

This is the right call given the constraints — there's no client build step, so the alternative is hand-copying these functions into the template string where they'd be untestable and would drift. And I checked it genuinely works today: the build is plain `tsc` (ES2022, no bundler/minifier), so the emitted source keeps identifiers intact, and all three functions currently reference only their parameters, locals, and JS built-ins.

The issue is durability. `Function.prototype.toString()` returns only the function's own body text — nothing it references from module scope travels with it. So this holds only while these stay self-contained, and nothing in `otp-input.ts` signals that. A maintainer there sees three ordinary exported pure functions with unit tests, and an entirely routine refactor breaks it:

```ts
const NON_ALPHANUMERIC = /[^A-Z0-9]/g   // hoist the regex

export function normalizeOtpValue(value, otpLength, otpCharset) {
  ...compactValue.toUpperCase().replace(NON_ALPHANUMERIC, '')
}
```

The serialized body now mentions `NON_ALPHANUMERIC`, which doesn't exist in the browser. Same for adding an import, extracting a helper, or switching to an arrow function that closes over module scope.

What makes it worth guarding is that nothing catches it:

- **`tsc` won't** — the server code still type-checks; the breakage lives inside a string.
- **Lint won't** — no rule models "this function is also a client bundle."
- **`otp-input.test.ts` won't** — it calls these in Node, where module scope *is* present, so it stays green.
- It surfaces as a `ReferenceError` in the browser on the OTP screen, where a JS error blocks sign-in outright.

Green CI, broken production login. Only Playwright would notice, and only if it exercises that screen.

Cheap fix — a test that evaluates each function's source in a scope with no module bindings, reproducing the browser's conditions:

```ts
// otp-input.test.ts
describe('client-serialized helpers', () => {
  // These are injected into the login page via toString(), so they must
  // reference only parameters, locals, and JS built-ins.
  it('stay self-contained when evaluated outside module scope', () => {
    const isolate = (fn: Function) =>
      new Function('return (' + fn.toString() + ')')()

    expect(isolate(normalizeOtpValue)('12 34 56', 6, 'numeric')).toBe('123456')
    expect(isolate(resolveOtpClickSelection)(3, 1)).toEqual({
      start: 1, end: 2, direction: 'forward',
    })
    expect(isolate(resolveOtpSelection)(4, 6, 3, 3, 4, 4)).toEqual({
      start: 3, end: 4, direction: 'backward',
    })
  })
})
```

`new Function` compiles in global scope, so a free identifier throws `ReferenceError` immediately. I verified both directions locally: the current functions pass isolated, and the hoisted-regex version above passes a normal call but throws `ReferenceError: NON_ALPHANUMERIC is not defined` when isolated — exactly the case that would otherwise reach production.

A short comment on the exports in `otp-input.ts` noting the constraint would help too, but the test is what actually holds the line since it doesn't rely on anyone reading it.

Separately: the underlying cause here is the ~580-line inline `<script>` that TypeScript and ESLint never see, which predates this PR. Filed as hypercerts-org/ePDS#213 — nothing for this branch to do about it.

.changeset/mobile-code-paste.md
Kzoeps: the breaking change is the styling and css targeting on the active code slot like it mentions above im checking to see if me  or gainforest uses this specific selector for css. Technically is a breaking change if anyone uses the styling mentioned above but ill make it backwards compatible since its just a single css that needs to be added

Kzoeps:
> I've tested this on mobile and it works for me, but it needs more clarity around whether it's a breaking change (it shouldn't be).
> 
> Also seems like a ton of new code for what I would have expected to be a pretty small tweak, but it seems that's because we're now taking over all control of handling input which was previously handled natively. I would have preferred to see it fixed whilst retaining native input control somehow, but maybe that's just not possible.

most of this code has been copied from input-otp which is what shadcn uses as well. previously we were using one input box for each otp slot and intercepting the paste functionality so that it pastes each number individually on each input:

how we are currently doing it right now:
`A paste handler called preventDefault(), split the pasted code, and assigned each character to successive inputs.`

Since this is now being changed to one single input and the slots are just visual hence the code changes. This is better interms of accessibility as well. and the code change is also due to adding support for clicking on otp input slots to change to that slot, arrows for changing the active slot.

.changeset/mobile-code-paste.md
Kzoeps: after analyzing it with maearths branding.css youre right its a breaking change for maearth interms of styling the otp. however to make it non breaking it would need to go back to single otp input boxes which i dont think is the way to do it.

ill change the release note to be a breaking change and not a patch

aspiers: I don't think that's the right way forward. We should try really hard to avoid breaking changes, otherwise the ecosystem devs will grow to hate us. In this case there is at least one way to achieve that, even if it's ugly.

aspiers: IOW, the code could simply check whether the branding CSS contains only the old selector and not the new one, and if so, automatically rewrite it.  It's really ugly, but it's WAY better than breaking every branded app out there.

At very least we need to offer a migration window; I would suggest a few months minimum. Another way is to simply maintain both the old and new mechanisms in parallel and auto-detect which one to use based on the branded CSS.  This is even uglier due to the duplication (albeit temporary), but arguably more correct.

aspiers:
> Since this is now being changed to one single input and the slots are just visual hence the code changes. This is better interms of accessibility as well. and the code change is also due to adding support for clicking on otp input slots to change to that slot, arrows for changing the active slot.

Thanks — that explains the size. Writing up the paste side in more detail for the record, since it took me a while to piece together from the diff, and it partly answers my own question about retaining native handling.

**There are two separate paths into the field, and only one of them is what #212 was about.**

*One-time-code autofill* — keyboard suggestion, SMS autofill — does **not** fire a `paste` event. It sets `.value` directly and fires `input`, so it is handled natively, then normalized by the `input` listener.

This is the actual #212 bug, and it never involved the paste handler at all: autofill wrote the whole code into one `maxlength="1"` field and the browser clipped it to a single character. No paste-handler change could have fixed that — consolidating to a single input is the fix, structurally. So on this path we *are* retaining native input control, which is what I was hoping for.

*Explicit paste* — Ctrl-V, iOS long-press → Paste — does fire `paste`, and is still intercepted:

```js
otpInput.addEventListener('paste', function(e) {
  if (!e.clipboardData) return;
  e.preventDefault();
  ...
```

What changed is what the handler does, not whether it runs. Before, it distributed characters across N inputs (`otpBoxes[idx + i].value = cleaned[i]`) and moved focus — the part you described as going away, which it has. Now it splices into one string, normalizes, restores the caret, and re-dispatches `input`; much less work, since the slots are just a render of `otpInput.value`.

This one can't be handed back to the browser, for the reason your code comment gives:

> Handle clipboard insertion before maxlength is applied so formatted values such as "123 456" can be normalized without the browser truncating the last character first.

`maxlength="6"` truncates *before* any listener runs, so `"123 456"` (7 chars) natively clips to `"123 45"` and the last digit is gone before `input` fires. Nothing downstream can recover it. That answers my "could this retain native input control" question for this path — not without dropping `maxlength`, which brings its own problems.

Related, and worth not losing: `opacity: 1` with transparent text (rather than `opacity: 0`) is load-bearing, not cosmetic — the real input has to stay hit-testable for iOS long-press paste to target it. Already commented in the CSS, and worth keeping that way.

Nothing to change here — just recording it, since "it's one input now, so paste works" undersells why both mechanisms are still needed, and could invite a later simplification pass to remove the handler.

.changeset/mobile-code-paste.md
Kzoeps: ill look into this. I was unable to think of a way to make it backwards compatible and couldnt come up with one. but what you said sounds reasonable and ill try it out!

packages/auth-service/src/routes/login-page.ts
Kzoeps: _(reply generated by OpenAI Codex)_

Fixed in `a7f9064`. `otp-input.test.ts` now serializes each injected helper with `toString()`, evaluates it through `new Function` outside module scope, and exercises the resulting function so future module-level dependencies fail CI.

Kzoeps:
_(reply generated by OpenAI Codex)_

Regarding [the latest CodeRabbit review](https://github.com/hypercerts-org/ePDS/pull/204#pullrequestreview-4854461533):

- Fixed the circular computed-style assertions in `b6bb55d`; the E2E step now checks five stable amber-theme signatures. This exposed and fixed missing `!important` declarations in the demo fixture.
- Fixed swallowed parser details in `b6bb55d`; the warning remains concise and the structured `err` is emitted at debug level.
- Verified that the installed selector parser accepts keyword and percentage keyframe offsets; added a regression covering the real caret keyframes rather than adding an unnecessary skip.
- Did not add synthetic `selector-parse-failed` / `serialization-failed` tests: the suggested keyframe case does not fail, and no stable public input was identified that reaches serialization failure without mocking implementation internals.
- Did not add tautological docstrings solely to satisfy the advisory percentage; the public adapter and invariant contract already have decision-focused documentation.

Kzoeps:
_(reply generated by OpenAI Codex)_

Regarding [the requested-changes review](https://github.com/hypercerts-org/ePDS/pull/204#pullrequestreview-4810210920): addressed in `a7f9064`. The single native OTP input remains, while common legacy input/focus/placeholder branding selectors are projected onto the visual slots automatically. Existing branded clients do not need a coordinated CSS migration for those forms, and the release note now describes the bounded compatibility contract.

Kzoeps:
@aspiers was working on the css parser to maintain backwards compatibility since duplicate inputs dont really work. 

However the css parser feels very fragile and also janky and adds a lot more code. Im leaning towards closing this pr and keeping the separate inputs and fixing the copy-paste only without the accessibility and navigation fixes.

One alternative here is to maybe have a field in the client metadata which is a field like `OtpInputMode: 'single-input' | 'slotted-input'` and based on that render the previous version or the new version.

let me know what you think

aspiers:
[**@Kzoeps**](https://github.com/Kzoeps) commented on [Aug 4, 2026, 14:51 GMT+1](https://github.com/hypercerts-org/ePDS/pull/204#issuecomment-5180009545 "2026-08-04T13:51:46Z - Replied by Github Reply Comments"):
> @aspiers was working on the css parser to maintain backwards compatibility since duplicate inputs dont really work. 
> 
> However the css parser feels very fragile and also janky and adds a lot more code. Im leaning towards closing this pr and keeping the separate inputs and fixing the copy-paste only without the accessibility and navigation fixes.

What parser is that?  If you really needed to parse CSS, I would say use treesitter, but that sounds overkill in this case since it's such a simple change needed - I would have thought a basic regexp search would do the job?  Probably easiest is search-and-replace, or it could automatically switch modes based on the search result.

> One alternative here is to maybe have a field in the client metadata which is a field like `OtpInputMode: 'single-input' | 'slotted-input'` and based on that render the previous version or the new version.

I'm not a fan of that because you would need it to default to the old version in order to achieve backwards compatibility, which would mean that every client wanting the new recommended version would have to explicitly request it.

If there is a concrete reason why a simple regexp search doesn't work, I'd like to discuss that before considering other options, because I have a fairly strong intuition that it should work.

Kzoeps:
> I would have thought a basic regexp search would do the job? Probably easiest is search-and-replace

Search and replace works but we need to know the exact css customization from each of the consumers. For now I have `maearth.com`, `gainforest.app`, `simocracy.org` and `certified.app` if we optimize for only these then a regexp works. The parser is the AI's more generic solution, i thought a .otp replace would work as well but upon looking at me's css they have css such as `input:focus` `input::placeholder` which would fail with a simple regex swap.

> I'm not a fan of that because you would need it to default to the old version in order to achieve backwards compatibility, which would mean that every client wanting the new recommended version would have to explicitly request it.

Makes sense.
 
> If there is a concrete reason why a simple regexp search doesn't work, I'd like to discuss that before considering other options, because I have a fairly strong intuition that it should work.

I think if we do it for the 4 apps mentioned above and based on their css write the regex then it would work. can switch it to that since i think the apps are maybe the only users of certified.one for now?

aspiers:
[**@Kzoeps**](https://github.com/Kzoeps) commented on [Aug 5, 2026, 08:55 GMT+1](https://github.com/hypercerts-org/ePDS/pull/204#issuecomment-5189133578 "2026-08-05T07:55:44Z - Replied by Github Reply Comments"):
> > I would have thought a basic regexp search would do the job? Probably easiest is search-and-replace
> 
> Search and replace works but we need to know the exact css customization from each of the consumers.

Why?  Doesn't it only have to replace `.otp-box:focus` with `.otp-box.active`?

> For now I have `maearth.com`, `gainforest.app`, `simocracy.org` and `certified.app` if we optimize for only these then a regexp works.

Probably that's good enough but we can survey all our trusted client brandings and check.

> The parser is the AI's more generic solution, i thought a .otp replace would work as well but upon looking at me's css they have css such as `input:focus` `input::placeholder` which would fail with a simple regex swap.

We don't need to touch more general selectors like `input:focus`.  Maybe you are talking about stuff like

```css
.otp-box {
    &:focus ...
```

?  What other combinations do we have to worry about in the absence of a parser?

> > If there is a concrete reason why a simple regexp search doesn't work, I'd like to discuss that before considering other options, because I have a fairly strong intuition that it should work.
> 
> I think if we do it for the 4 apps mentioned above and based on their css write the regex then it would work. can switch it to that since i think the apps are maybe the only users of certified.one for now?

Bit more than just those 4 but the nice thing is that we have the exact list of trusted clients so we can do a 100% reliable check!  Would you be able to do that?

PR #215, opened 2026-07-30
P1: fix(auth): clear email when returning from OTP entry

aspiers:
## Summary

Make **Use different email** start from a clean email field rather than retaining the previous account. The field is cleared and focused so the next keystroke goes to the intended place.

## Changes

- Clear and focus the email field when returning from the OTP step
- Cover the behavior with unit and Cucumber assertions

## Testing

- `pnpm format:check`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test`
- `pnpm test:coverage`

## Screenshots

**Before:** **Use different email** returned to a form that retained the previous address.

Before: email address retained

**After:** the form returns with a blank, focused email field.

After: cleared email form

## Notes

- Deployed E2E suite passed against the PR #215 Railway environment in [run 30549079555](https://github.com/hypercerts-org/ePDS/actions/runs/30549079555).

### DIFF
--- a/.changeset/use-different-email-clears-form.md
+++ b/.changeset/use-different-email-clears-form.md
@@ -1,0 +1,9 @@
+---
+'ePDS': patch
+---
+
+"Use different email" on the sign-in code page now clears the email field.
+
+**Affects:** End users
+
+**End users:** clicking **Use different email** on the code-entry page used to take you back to the email form with your previous address still filled in — exactly the opposite of what the button suggests, and forcing you to clear the field manually before you could type a new address. The form now starts empty, with the cursor in the field, so you can just type the new address and continue.
--- a/e2e/step-definitions/auth.steps.ts
+++ b/e2e/step-definitions/auth.steps.ts
@@ -929,3 +929,14 @@ Then(
     })
   },
 )
+
+// ---------------------------------------------------------------------------
+// "Use different email" UX
+// ---------------------------------------------------------------------------
+
+Then('the email input is empty and focused', async function (this: EpdsWorld) {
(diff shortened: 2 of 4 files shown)


aspiers:
The PR-attached E2E check is red only because its SHA-specific Railway deployment lookup timed out before the preview deployment was discoverable. The same commit passed the full deployed suite when run explicitly against `ePDS / ePDS-pr-215`: https://github.com/hypercerts-org/ePDS/actions/runs/30549079555

packages/auth-service/src/__tests__/login-page.test.ts
```diff
@@ -627,6 +627,12 @@ describe('renderLoginPage inline Resend action on expired OTP', () => {
       /if \(isExpired\) \{[\s\S]*?\} else \{[\s\S]*?showError\(result\.error\);\s*\}/,
     )
   })
+
+  it('clears and focuses the email input when returning from the OTP step', () => {
+    const html = renderDefault()
+    expect(html).toMatch(/emailInput\.value\s*=\s*''/)
+    expect(html).toMatch(/emailInput\.focus\(\)/)
```
aspiers: Completely unacceptable approach to testing as mentioned in another PR just now!

aspiers: Fixed in 1a90062. I removed the rendered-JavaScript/source-fragment test. The regression is now covered by the browser scenario, which clicks “Use different email” and asserts that the live email input is empty and focused.

aspiers:
@coderabbitai review

PR #216, opened 2026-07-30
P0: fix(auth): ignore incomplete OTP submissions

aspiers:
## Summary

Stop empty or partial verification codes before they reach better-auth. This avoids misleading invalid-code errors and prevents incomplete submissions from spending a verification attempt.

**Verify** is disabled until every OTP slot is filled, so an incomplete code is visibly un-submittable rather than a click that silently does nothing. The submit-time guard remains as a backstop for the paths that reach submit without the button (Enter, autofill, `requestSubmit()`).

## Changes

- Disable **Verify** until every OTP slot is filled, driven by the filled-slot count
- Keep the submit-handler completeness guard as a backstop
- Focus the first unfilled slot when a bypassing submit is ignored
- Never re-enable **Verify** mid-verify or after the flow is aborted — including on the failed-verify path, which clears the boxes
- Re-sync at load so a bfcache-restored page can't strand the button disabled
- Fade the disabled button to 0.6 opacity (reads as inert; keeps the label at ~4.9:1, WCAG AA) and stop `:hover` brightening it
- Add unit and Cucumber coverage

## Testing

- `pnpm format:check`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test`
- `pnpm test:coverage`

## Screenshots

**Before:** submitting a partial code sent it for verification and showed a misleading invalid-code error. The preview response was controlled to isolate this state.

Before: partial OTP reported as invalid

**After:** clicking **Verify** with only part of the code entered sends no verification request, leaves the controls editable, and shows no false error.

After: incomplete OTP remains editable

## Notes

- Rebased onto current `main` at a6cf09d; conflicts with the newer “Use different email” coverage were resolved by retaining both behaviors.
- The completeness guard now applies cleanly to the segmented OTP boxes already on `main`.
- Current rebased SHA aa1abc7 passed formatting, lint, typecheck, production build, unit tests, coverage, Cucumber definition validation, and the full deployed E2E suite in [run 30652050690](https://github.com/hypercerts-org/ePDS/actions/runs/30652050690).
- Follow-up commit c0f0867 disables the button (previously the guard was submit-time only, leaving Verify enabled but inert). All 17 checks green on c0f0867; the deployed E2E suite ran 73 scenarios / 494 steps, all passing, including both new `@verify-incomplete-otp` scenarios.

### DIFF
--- a/features/passwordless-authentication.feature
+++ b/features/passwordless-authentication.feature
@@ -403,31 +403,36 @@ Feature: Passwordless authentication via email OTP
     When the user clicks "Use different email"
     Then the email input is empty and focused
 
-  # Clicking Verify before the code was complete used to flash "Invalid OTP",
+  # Submitting before the code was complete used to flash "Invalid OTP",
   # which is dishonest: the user had not entered an invalid full code. It also
   # burned a real call to better-auth's /sign-in/email-otp endpoint, which
-  # counts against the rate-limiter. Cover both the empty and partial states.
+  # counts against the rate-limiter. Verify is now disabled until every slot
+  # is filled, and the submit handler still guards the paths that bypass the
+  # button (Enter, autofill). Cover both the empty and partial states.
   @email @verify-incomplete-otp
-  Scenario: Submitting Verify with an empty code does not send a verification request
+  Scenario: Verify is disabled and an empty submit sends no verification request
     When the demo client initiates an OAuth login
     Then the browser is redirected to the auth service login page
     And the login page displays an email input form
     When the user enters a unique test email and submits
     Then the login page shows an OTP verification form
-    When the user clicks the Verify button without entering a code
+    And the Verify button is disabled
+    When the user tries to submit the OTP form without entering a code
     Then no OTP verification request is sent
     And no "Invalid OTP" error is shown
+    And the Verify button is disabled
 
   @email @verify-incomplete-otp
-  Scenario: Submitting Verify with a partial code does not send a verification request
+  Scenario: Verify stays disabled and a partial submit sends no verification request
     When the demo client initiates an OAuth login
     Then the browser is redirected to the auth service login page
     And the login page displays an email input form
     When the user enters a unique test email and submits
     Then the login page shows an OTP verification form
-    When the user enters two OTP digits and clicks the Verify button
+    When the user enters two OTP digits and tries to submit the OTP form
     Then no OTP verification request is sent
     And no "Invalid OTP" error is shown
+    And the Verify button is disabled
 
   @email @demo-cookie-expiry @bug-report
   Scenario: Demo client's OAuth cookie has expired by the time of callback — useful error, not generic auth_failed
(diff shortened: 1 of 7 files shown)


packages/auth-service/src/__tests__/login-page.test.ts
```diff
@@ -627,6 +627,11 @@ describe('renderLoginPage inline Resend action on expired OTP', () => {
       /if \(isExpired\) \{[\s\S]*?\} else \{[\s\S]*?showError\(result\.error\);\s*\}/,
     )
   })
+
+  it('short-circuits Verify until every OTP slot is filled', () => {
+    const html = renderDefault()
+    expect(html).toMatch(/if \(otp\.length < otpBoxes\.length\)/)
```
aspiers: WTF, checking that client-side code fragments are present is not a valid test! You need to check the actual behaviour, not just blindly hope that the presence of some code implements it correctly!! I only just noticed that you have been continuing this horrendous anti-pattern which already exists elsewhere in this file. This is very misguided and unacceptable AI agent behaviour which needs fixing.

features/passwordless-authentication.feature
```diff
@@ -384,6 +384,23 @@ Feature: Passwordless authentication via email OTP
...
+  # rate-limiter. The Verify button must not submit until the full
+  # code has been entered.
+  @email @verify-empty-otp
+  Scenario: Submitting Verify with an empty code does not produce a misleading "Invalid OTP"
+    When the demo client initiates an OAuth login
+    Then the browser is redirected to the auth service login page
+    And the login page displays an email input form
+    When the user enters a unique test email and submits
+    Then the login page shows an OTP verification form
+    When the user clicks the Verify button without entering a code
```
aspiers: This misses the case where only a partial code is entered.  That should be covered at least by functional tests if not E2E.

aspiers (commented):
The intention is good here but very flawed testing.

aspiers:
The PR-attached E2E check is red only because its SHA-specific Railway deployment lookup timed out before the preview deployment was discoverable. The same commit passed the full deployed suite when run explicitly against `ePDS / ePDS-pr-216`: https://github.com/hypercerts-org/ePDS/actions/runs/30549081436

packages/auth-service/src/__tests__/login-page.test.ts
aspiers: Fixed in ce5653a. I removed the generated-JavaScript/source-fragment assertion entirely. The regression is now covered behaviorally in Playwright/Cucumber: submitting an incomplete OTP is exercised in the browser, the test records requests to `/sign-in/email-otp`, and asserts that no verification request or misleading error occurs. I also audited the other split PR additions and removed the same anti-pattern there.

features/passwordless-authentication.feature
aspiers: Fixed in ce5653a. The feature now covers both zero entered digits and a partially entered code. In each case it clicks Verify, asserts no `/sign-in/email-otp` request was sent, and confirms the editable OTP form remains without a false error.

aspiers:
_(reply generated by OpenAI Codex)_

Closing and reopening the PR created a fresh Railway preview for the unchanged current SHA. The required check has now run the full E2E suite and passed: https://github.com/hypercerts-org/ePDS/actions/runs/30557766403

PR #218, opened 2026-07-30
P0: fix(auth): preserve the request URI in recovery links

aspiers:
## Summary

Carry the active PAR `request_uri` into account recovery instead of using a placeholder. Successful recovery can then return the user to the authorization request that initiated sign-in.

## Changes

- Render recovery links with the active request URI
- Thread the value through regular and preview render paths
- Cover the round trip with unit and Cucumber assertions

## Testing

- `pnpm format:check`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test`
- `pnpm test:coverage`

## Before/after evidence

**Before:** the recovery link used a placeholder `request_uri`. Clicking **Back to sign in** therefore could not return to the active authorization request or restore its OTP form.

**After:** the browser scenario now captures the live `request_uri`, follows the recovery link, clicks **Back to sign in**, verifies the same URI is restored, and asserts that a usable sign-in form is visible instead of an OAuth error. This is a navigation/state fix with no meaningful static visual difference; the complete behavioral round trip is covered in [`account-recovery.feature`](https://github.com/hypercerts-org/ePDS/blob/79ca2e2384801ca9708c312e08acfe317824dfbc/features/account-recovery.feature#L43-L52).

## Notes

- The earlier PR-attached E2E attempt could not find the delayed deployment for the final amendment. After Railway recovered, the required check was rerun and the current SHA passed the full deployed suite in [run 30561186061](https://github.com/hypercerts-org/ePDS/actions/runs/30561186061).

- Current-branch deployed E2E: [run 30561478453](https://github.com/hypercerts-org/ePDS/actions/runs/30561478453).

- The deployed E2E suite [passed](https://github.com/hypercerts-org/ePDS/actions/runs/30547814692).


features/account-recovery.feature
```diff
@@ -39,6 +39,15 @@ Feature: Account recovery via backup emails
...
     And no email arrives for that non-existent address
 
+  # The recovery link must carry the active authorization request rather
+  # than a placeholder, otherwise Back to sign in cannot round-trip.
+  @recovery-link-roundtrip
+  Scenario: Recovery link carries the active OAuth flow's request_uri
+    Given the demo client initiates OAuth with the test email as login_hint
+    Then an OTP email arrives in the mail trap
+    And the login page shows an OTP verification form
+    Then the recovery link points at the active OAuth flow's request_uri
```
aspiers: This doesn't seem like it's covering the full round trip?  The changeset says "hitting **Back to sign in** at the end" but this only goes as far as checking the recovery link is correct.

aspiers: Fixed in a193589. The scenario now performs the complete browser round trip: it captures the active `request_uri`, follows Recover with backup email, clicks Back to sign in, verifies the returned URL carries the same `request_uri`, and asserts the original OTP form is restored. I also removed the old rendered-HTML string assertion so this is covered by behavior rather than source inspection.

aspiers: Follow-up from deployed E2E: the full round trip correctly returns to the same live `request_uri`, but PAR hint lookup is not guaranteed to be repeatable, so the returning page may legitimately show either the email or OTP step. Final commit b9bfa78 now asserts the actual contract: same active request, a visible sign-in form, and no OAuth error page. Current-branch E2E run: https://github.com/hypercerts-org/ePDS/actions/runs/30560775388.

aspiers: Final follow-up: commit 79ca2e2 fixes the Playwright assertion to count exactly one visible sign-in form (rather than applying a strict visibility assertion to both DOM forms). Current-branch E2E: https://github.com/hypercerts-org/ePDS/actions/runs/30561181344.

aspiers: The duplicate manually dispatched run was cancelled to avoid two suites competing for the same Mailpit/Railway environment. The final PR-attached E2E run is https://github.com/hypercerts-org/ePDS/actions/runs/30561186061.

aspiers: Correction: because 79ca2e2 changes only E2E files, the PR-attached workflow has no corresponding Railway deployment and remains in deployment discovery. The final explicit environment run is https://github.com/hypercerts-org/ePDS/actions/runs/30561478453.

Issue #214, opened 2026-07-30
Flaky e2e: "Expired OTP is rejected" reports "Invalid OTP" instead of "OTP expired"


aspiers:
## Summary

The e2e scenario `Expired OTP is rejected, resend recovers the flow` (`features/passwordless-authentication.feature:257`, tagged `@email @otp-expiry`) fails intermittently:

```
Then the verification form shows an "OTP expired" error
  Error: expect(locator).toContainText(expected) failed
  Locator: locator('#error-msg')
  Expected substring: "OTP expired"
  Received string:    "Invalid OTP"
  Timeout: 10000ms
```

Observed on PR #148 (run [30533229391](https://github.com/hypercerts-org/ePDS/actions/runs/30533229391), commit `1104efa`): 71 of 72 scenarios passed, this one failed. Re-running the identical commit passed with no code change, confirming flake rather than regression. The accompanying `page.click` / `page.screenshot` timeouts in that log are Cucumber retry noise from the same scenario, not independent failures.

## Why the symptom is misleading

`"Invalid OTP"` does **not** mean a wrong code was submitted. In better-auth 1.4.18 (`dist/plugins/email-otp/routes.mjs:394-396`) the sign-in verify path is:

```js
const verificationValue = await ctx.context.internalAdapter
  .findVerificationValue(`sign-in-otp-${email}`);
if (!verificationValue) throw new APIError("BAD_REQUEST", { message: ERROR_CODES.INVALID_OTP });
if (verificationValue.expiresAt < new Date()) throw new APIError("BAD_REQUEST", { message: ERROR_CODES.OTP_EXPIRED });
```

So `INVALID_OTP` is thrown when the lookup returns **no row at all**, and `OTP_EXPIRED` only when a row is found *and* is past expiry. Our own code already documents this ambiguity — see the comment above `OTP_FAILURE_MESSAGES` in `packages/auth-service/src/better-auth.ts`:

> Note on "Invalid OTP": better-auth throws it both for a wrong code and for no pending code at all

The scenario therefore fails whenever the verification row is missing rather than merely stale.

## Suspected cause: the test hook lacks a liveness guard

`POST /_internal/test/expire-otp` in `packages/auth-service/src/routes/test-hooks.ts:118`:

```sql
UPDATE verification SET expiresAt = ? WHERE identifier = ?
```

Compare the sibling `expire-auth-flow` hook in the same file (lines 67 and 72), which **does** constrain to live rows:

```sql
[...]

PR #220, opened 2026-07-30
P1: fix(auth): improve OTP resend UX

aspiers:
## Summary

Clear the previously entered verification code after a successful resend. The fresh code now starts with an empty control focused at its first position.

## Changes

- Reset OTP state after resend succeeds
- Focus the first entry position
- Add focused render coverage

## Testing

- `pnpm format:check`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test`
- `pnpm test:coverage`

## Screenshots

**Before:** a successful resend left stale digits in the OTP boxes.

Before: resend retains stale digits

**After:** the old digits are cleared and focus returns to the first box. (This shot predates the later commits on this branch, so its banner shows the older copy and position; both are covered below.)

After: cleared OTP after resend

### Errors at the point of failure

Captured against live Railway previews — PR #229's for the before, which still serves the old layout, and this branch's own for the after.

**Before:** a rejected code reported itself above the subtitle, several elements from the boxes the user had just typed into, with no way forward but the standalone **Resend code** button further down.

Before: error above the form

**After:** the message sits between the boxes and **Verify**, where the user is already looking. Because this particular failure invalidates the stored code server-side, it also carries an inline **Resend code** action — there is nothing left to retype. A plain mistyped code keeps the message without the action, since retyping is the right recovery there.

After: error at the point of failure

### Resend confirmation

**After:** the confirmation acknowledges the click, then warns that earlier codes have stopped working and points at the spam folder — the two things most likely to be wrong for anyone who got as far as resending.

After: resend confirmation

## Notes

- Current-branch deployed E2E: [run 30560777504](https://github.com/hypercerts-org/ePDS/actions/runs/30560777504).

- PR #204 introduces a single-input OTP controller; whichever PR lands second should retain this behavior through its `clearOtpBoxes()` helper.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resending a sign-in code now clears entered digits and refocuses the first code field.
  * Users are informed that only the newest code is valid and should check their spam folder.
  * Sign-in errors appear beside the relevant email or verification-code field.
  * Invalidated or excessive-attempt codes offer an inline resend option.

* **Tests**
  * Added coverage for resetting the verification form during expired-code recovery.


packages/auth-service/src/__tests__/login-page.test.ts
```diff
@@ -627,6 +627,13 @@ describe('renderLoginPage inline Resend action on expired OTP', () => {
       /if \(isExpired\) \{[\s\S]*?\} else \{[\s\S]*?showError\(result\.error\);\s*\}/,
     )
   })
+
+  it('clears and refocuses the OTP boxes after a successful Resend', () => {
+    const html = renderDefault()
+    expect(html).toMatch(
+      /if \(result\.error\) \{[\s\S]*?\} else \{[\s\S]*?clearOtpBoxes\(\);[\s\S]*?otpBoxes\[0\]\.focus\(\);[\s\S]*?showSuccess\('Code resent!'\)/,
```
aspiers: Completely unacceptable approach to testing! as per https://github.com/hypercerts-org/ePDS/pull/216#discussion_r3683648403

aspiers: Fixed in 50f4a5e. I removed the generated-client-script assertion. The resend behavior is now exercised in Playwright/Cucumber: the browser enters old digits, performs a successful resend, and verifies that all boxes are cleared and focus returns to the first box.

aspiers: Follow-up from deployed E2E: the behavioral assertion initially assumed the test config’s 8-digit length while the preview environment issued 6 digits. Final commit 619ae08 derives the expected box count from the fresh OTP actually received, then verifies every box is empty and the first is focused. Current-branch E2E run: https://github.com/hypercerts-org/ePDS/actions/runs/30560777504.

aspiers:

@coderabbitai re: the nitpick on `e2e/step-definitions/auth.steps.ts` (the "assert the old digits before testing the reset" suggestion) — agreed and applied in d73d017.

Your reasoning is right: without it, a regression that cleared or ignored input *before* the resend would still satisfy the final empty-state assertion, so the scenario wouldn't actually be proving the reset.

```ts
await otpBoxes.nth(0).fill('1')
await otpBoxes.nth(1).fill('2')
// Prove the digits actually landed, so the later empty-box assertion
// demonstrates that resend cleared them rather than that they were
// never entered.
await expect(otpBoxes.nth(0)).toHaveValue('1')
await expect(otpBoxes.nth(1)).toHaveValue('2')
```

aspiers:

Re: the [Blacksmith report](https://github.com/hypercerts-org/ePDS/pull/220#issuecomment-5133372222) of `Cucumber/Expired OTP is rejected, resend recovers the flow` failing — that result is **stale**, and for the record it was a real bug rather than the known flake in #214.

It ran against 50f4a5e9, an earlier version of the commit that has since been amended and force-pushed away (it is no longer an ancestor of this branch). The failure was:

```
Error: expect(locator).toHaveCount(expected) failed
  Expected: 8
  - unexpected value "6"
```

**Cause:** my new step asserted `toHaveCount(testEnv.otpLength)`. `e2e/support/env.ts:33` defaults `otpLength` to `8` when `OTP_LENGTH` is unset in the e2e runner's environment, but the auth service under test renders 6 boxes — so the assertion compared an env-derived default against the real DOM.

**Fix**, already in 619ae08: assert against the length of the OTP actually captured from the mail trap, which is by definition the number of boxes to expect:

```ts
await expect(otpBoxes).toHaveCount(this.otpCode.length)
```

This mirrors the existing guard at `auth.steps.ts:55`, which likewise prefers the live box count over `testEnv.otpLength`.

E2E tests pass on 619ae08 ([run 30560777504](https://github.com/hypercerts-org/ePDS/actions/runs/30560777504)), and all checks are currently green on the branch tip.

packages/auth-service/src/routes/login-page.ts
```diff
@@ -1324,7 +1324,18 @@ export function renderLoginPage(opts: {
...
+          // Clear any characters typed for the old code so the new code
+          // starts from a clean, focused input grid.
+          clearOtpBoxes();
+          if (otpBoxes.length) otpBoxes[0].focus();
+          // Both facts only matter once a resend has happened, so they
+          // live here rather than in permanently-visible page copy:
+          // sending a new OTP invalidates every earlier one, and a user
+          // who needed to resend is the user whose mail may be in spam.
+          showSuccess(
+            'Use the new code; earlier ones no longer work. ' +
```
aspiers:
```suggestion
            'Resent! Make sure to use the new code; earlier ones no longer work. ' +
```

aspiers: _(reply generated by Claude Opus 5 via Claude Code)_

Applied in e9977ab.

Better than what I had. I dropped `Code resent!` on the grounds that it restated what the click already confirmed, but that reasoning was wrong: the button reverts to its label immediately, so without an acknowledgement the user is left inferring from a warning alone that anything happened. Leading with "Resent!" confirms first, then warns.

packages/auth-service/src/routes/login-page.ts
```diff
@@ -559,6 +559,14 @@ export function renderLoginPage(opts: {
   const hasGithub = 'github' in socialProviders
   const hasSocialProviders = hasGoogle || hasGithub
 
+  // The flash region is a single element shared by both steps, so there
+  // is only ever one live region for assistive tech to track. It is
+  // server-rendered into whichever step is initially visible and moved
+  // between the two slots on step transitions; both transitions call
+  // clearError() first, so it is always empty when it moves.
+  const flashRegionHtml =
+    '<div id="error-msg" class="flash-msg" style="display:none;" role="status" aria-live="polite"></div>'
```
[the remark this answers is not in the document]
aspiers: _(reply generated by Claude Opus 5 via Claude Code)_

Applied in e9977ab, as suggested: `.flash-msg.hidden { display: none; }` in the stylesheet, `classList.remove(_hidden_)` in `setFlash()`, `classList.add(_hidden_)` in `clearError()`.

Worth recording that the inline style was **pre-existing** — it came from the original `#error-msg` markup, and this PR only moved that markup between the two step containers. So the finding is a genuine cleanup rather than a regression introduced here. Fixing it in this commit anyway, since relocating the element is the natural point to do it and leaving it would carry the inconsistency into a new location.

The codebase corroborates the guideline independently of the config: `.step-otp.active` and `.step-email.hidden` already gate visibility by class, so the region was the odd one out.

I also reworded a nearby comment in `clearError()` that referred to "the `display:none`" as a concrete thing; it now refers to the region being hidden, which stays accurate under class-based control.

docs/screenshots/pr-220-before-error-above-form.png
aspiers, line 1: These do not belong in the repo, only in the PR description.

PR #227, opened 2026-07-30
P1: fix(auth): use truthful unavailable-handle copy

aspiers:
## Summary

Describe rejected handles as unavailable rather than always claiming that another user already took them. This keeps the UI truthful for reserved names and other policy rejections.

## Changes

- Normalize live-check and submit-time copy to **Not available**
- Add focused handle-picker render coverage
- Document the end-user change with a changeset

## Testing

- `pnpm format:check`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test`
- `pnpm test:coverage`

## Screenshots

**Before:** the unavailable state overclaimed that the handle was already taken.

Before: handle picker overclaims already taken

**After:** it uses truthful generic copy. The preview responses were controlled to isolate both wording states.

After: unavailable handle copy

## Notes

- Focused extraction and review of work originally proposed in #165.
- The reserved-handle availability PR is stacked on this branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated handle picker messaging to clearly state “Not available.” for reserved or unavailable handles.
  * Removed misleading “Already taken” wording when a handle is permanently unavailable.
  * Unavailable handle states now display consistently across themed interfaces.

* **Tests**
  * Added automated coverage verifying the updated availability message and preventing incorrect wording from appearing.

PR #228, opened 2026-07-30
P1: fix(pds): reject reserved handles in live checks

aspiers:
## Summary

Make the live handle-availability endpoint apply the same reserved-name policy as account creation. Users no longer see **Available** for a handle that submission will immediately reject.

## Changes

- Reuse upstream handle normalization and service constraints
- Return unavailable only for the upstream `HandleNotAvailable` policy error
- Add focused reserved-handle tests and a changeset

## Testing

- `pnpm format:check`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test`
- `pnpm test:coverage`

## Screenshots

Not applicable: this is server-side handle-policy enforcement. It reuses the existing **Not available** picker state from #227 and adds no new rendering.

## Notes

- The PR-attached E2E deployment lookup is expected to fail while this remains stacked: Railway does not create a standalone preview environment for the stacked head. The agreed plan is to wait for #227 to merge, then retarget/rebase this PR and run the full deployed suite before merge.

- This is a stacked PR based on `split-pr165/handle-unavailable-copy`.
- The upstream helper is imported through the version-pinned `@atproto/pds/dist/handle/index.js` path because it is not exported from the package root.

### DIFF
--- a/.changeset/handle-picker-flags-reserved-handles.md
+++ b/.changeset/handle-picker-flags-reserved-handles.md
@@ -1,0 +1,9 @@
+---
+'ePDS': patch
+---
+
+Reserved handles are now shown as unavailable before account creation.
+
+**Affects:** End users
+
+**End users:** the live handle check no longer labels names reserved by the service, such as `admin`, `support`, or `www`, as available only to reject them after submission. The picker reports them as unavailable immediately so you can choose another name without losing progress.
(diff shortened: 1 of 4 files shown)


aspiers:
The E2E check stops at **Find Railway deployment** before running the suite. This PR is intentionally stacked on #227, so Railway did not create a standalone PR #228 preview environment. Local format, lint, typecheck, unit tests, and coverage all pass; the failure is preview-environment availability rather than a test assertion.

PR #232, opened 2026-08-03
fix(auth): darken muted sign-in text to meet WCAG AA

aspiers:
## What

Changes `--muted-foreground` on the sign-in page from `#999` to `#666`, and routes `.divider` through that variable instead of its own hard-coded `#999`.

## Why

`--muted-foreground` carries the 13px terms line, the "Recover with backup email" link, the "or continue with" / "or use email" separators, and the "powered by Certified" footer. Measured against the two backgrounds those elements actually sit on:

| | card `#F8F8F8` | page `#E8E8E8` |
|---|---|---|
| `#999` (current) | 2.68:1 ❌ | 2.33:1 ❌ |
| `#6b6b6b` | 5.02:1 ✅ | 4.35:1 ❌ |
| `#666` (this PR) | 5.41:1 ✅ | 4.69:1 ✅ |

Both current values fall far below the 4.5:1 WCAG AA minimum for normal-size text — and below even the 3.0:1 large-text threshold, which does not apply at 13px anyway.

## Screenshots

Both are real Railway preview deployments, not local mocks. The OTP step is shown so three consumers of the variable are in frame at once.

**Before:** muted text at `#999` — washed out against the card, and worse against the page background behind the footer.

Before: low-contrast muted text

**After:** the same elements at `#666`, clearing AA on both backgrounds.

After: muted text meets WCAG AA

Computed colours at capture time confirm the swap reached every consumer: `rgb(153,153,153)` before, `rgb(102,102,102)` after, for `.terms`, `.recovery-link` and `.powered-by` alike.

## Why `#666` and not `#6b6b6b`

`#6b6b6b` was the value proposed alongside the spam-folder hint in #221, but it clears AA only against the card. `.powered-by` lives in `.page-wrap` *outside* `.container`, so it renders against the page background — visible in the screenshots above as the lighter band below the card — where `#6b6b6b` reaches just 4.35:1 and still fails at 13px. `#666` passes on both and is a conventional shorthand.

## The `.divider` follow-up

Copilot caught that `.divider` ("or continue with", "or use email") hard-coded `color: #999` at 13px rather than using the variable, so darkening `--muted-foreground` would have left it behind at 2.68:1 — the same defect, in the one place that had opted out of the mechanism being fixed. Fixed in 6ed3d05.

Side effect worth recording: the divider now honours a `branding.css` override of `--muted-foreground`, which it previously ignored. Noted in the changeset's operator section. It does not appear in the screenshots because dividers only render when social providers are configured.

## Provenance

Extracted from #221, where it rode along as an undocumented global change justified by a hint element that PR was adding. That hint is superseded by a resend-time message in #220, so the original motivation is gone — but the contrast defect predates #221 and stands on its own.

## Testing

`format`, `lint`, `typecheck` all pass; 1084 unit tests pass. Contrast ratios computed with the WCAG 2.x relative-luminance formula.


aspiers:
> the 13px `.divider` text is still hard-coded to `#999`

Good catch — fixed in 6ed3d05.

`.divider` ("or continue with", "or use email") opted out of the variable and hard-coded `color: #999` at 13px, so darkening `--muted-foreground` left it behind at 2.68:1 against the card. That is exactly the defect this branch exists to fix, in the one place that had opted out of the mechanism being fixed. Both dividers render inside `.container`, so the card ratio is the relevant one.

It now uses `var(--muted-foreground)`. One side effect worth recording: the divider will now honour a `branding.css` override of `--muted-foreground`, which it previously ignored — noted in the changeset's operator section.

.changeset/muted-text-contrast-on-sign-in.md
```diff
@@ -0,0 +1,11 @@
...
+'ePDS': patch
+---
+
+Small grey text on the sign-in page is darker and easier to read.
+
+**Affects:** End users, Operators
+
+**End users:** the 13px grey text — the terms line, the "Recover with backup email" link, the "or continue with" / "or use email" separators, and the "powered by Certified" footer — was too light to meet the WCAG AA contrast minimum against either the card or the page background behind it. It is now dark enough to pass on both.
+
+**Operators:** the `--muted-foreground` custom property on the sign-in page changes from `#999` to `#666`. If your `branding.css` overrides it, check your value clears 4.5:1 against both `#F8F8F8` (the card) and `#E8E8E8` (the page); the old default reached only 2.68:1 and 2.33:1 respectively. The separator text between the social and email sign-in options now follows `--muted-foreground` too, where it previously ignored the override and stayed grey.
```
aspiers: It's devs of trusted client app who author `branding.css`, isn't it? Or at least those devs certainly need to think about this if they use custom branding, whether operators do or not.

aspiers: _(reply generated by Claude Opus 5 via Claude Code)_

You are right, and it was miscategorised. Fixed in 9b6b1f6 — reattributed from Operators to Client app developers.

`branding.css` is a field in the client's own `client-metadata.json`, not an operator surface. `docs/configuration.md` describes it as what "trusted clients can provide", `docs/tutorial.md` addresses it to "trusted clients" in the client tutorial, and every prior CHANGELOG entry mentioning it files it under **Client app developers**. An operator's only involvement is gating which `client_id`s may inject CSS at all, via `PDS_OAUTH_TRUSTED_CLIENTS` — they never author the file, so the adaptation instruction was pointed at a reader who cannot act on it.

Reworded to say "your override" rather than a third party's, and I added the point that actually matters for anyone who did set the property: a value picked to sit alongside the old `#999` may itself be below AA, so it is worth rechecking rather than assuming the default change is the whole story.

To your "or at least those devs certainly need to think about this" — agreed, and that is now the only audience listed alongside End users. I considered keeping Operators as a secondary audience on the grounds that they choose the trust allowlist, but that felt like noise: nothing in this change alters what an operator configures.

PR #240, opened 2026-08-04
ci(coveralls): add delta tolerance and absolute floor

aspiers:
Split out of #80, which bundled this CI-policy change with an unrelated one-line security fix (redacting `request_uri` from a warn log). That fix has been waiting since April; separating them lets it land while this is settled on its own merits.

## What it does

Adds `.coveralls.yml`:

```yaml
coverage_decreased_threshold: 0.05
coverage_threshold_for_failure: 32.3
```

Intent: absorb measurement jitter with a 0.05% per-PR drop tolerance, and set a 32.3% absolute floor to stop unbounded ratchet drift.

## ⚠️ These key names are probably wrong

Raised by CodeRabbit on #80. Investigating it produced a more nuanced answer than either of us started with:

**Thresholds *can* be set in `coveralls.yml`.** A Coveralls maintainer says so in [lemurheavy/coveralls-public#1455](https://github.com/lemurheavy/coveralls-public/issues/1455) — so "these are web-UI-only settings" is not correct. But under a **nested** structure:

```yaml
status:
  change_threshold: 5
  api:
    fail_threshold: 90
```

**The keys here match neither that shape nor the API field names** (`commit_status_fail_threshold`, `commit_status_fail_change_threshold`). A GitHub code search finds **zero** other repositories using `coverage_threshold_for_failure` in a `.coveralls.yml`, against a real hit for the nested `change_threshold` form.

The upstream evidence is itself weak: the maintainer offered that snippet untested, and the reporter closed the loop with *"my co-worker just changed the values in the web UI. So I never tested this, sorry"*. The issue was closed on an unverified suggestion.

**Not confirmed either way.** `GET /api/repos/github/hypercerts-org/ePDS` returns 403 without a Coveralls token, and the public `.json` endpoint exposes build data only — no threshold settings. Anyone with the token can settle it by comparing the configured thresholds against this file.

## Before merging

Either confirm the flat keys work, or switch to the nested form and verify. As it stands this file may configure nothing while appearing to set policy — which is worse than not having it, since it invites the belief that thresholds are version-controlled when they are still whatever the web UI says.

### DIFF
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,0 +1,2 @@
+coverage_decreased_threshold: 0.05
+coverage_threshold_for_failure: 32.3


aspiers:
Closing in favour of #245.

@copilot-pull-request-reviewer was right: these YAML keys configure nothing. Verified:

- The Coveralls docs list no such keys — `docs.coveralls.io/configuration` and `/coveralls-yaml` both 404, and the only documented `.coveralls.yml` key anywhere on the docs site is `repo_token`.
- `coveralls-ruby`'s `lib/coveralls/configuration.rb` parses exactly three keys: `repo_token`, `repo_secret_token`, `service_name` (Travis only).
- Decisively for us: `.github/workflows/ci.yml` uses `coverallsapp/github-action@v2`, whose documented inputs include no threshold option and no `.coveralls.yml` reading. Its only failure input, `fail-on-error`, governs upload failures rather than coverage metrics — so the file was never transmitted to Coveralls in the first place.
- `docs.coveralls.io/app-ui` confirms "Coverage threshold" and "Decrease threshold" as Repo Settings fields, and recurring upstream requests (lemurheavy/coveralls-public#1616, #1106) confirm no YAML equivalent exists.

So the file would have looked like version-controlled policy while doing nothing — Copilot's option (c), roughly.

The thresholds have now been set in the Coveralls dashboard instead:

| Setting | Value |
| --- | --- |
| Coverage threshold for failure | 55% |
| Coverage decrease threshold for failure | 0.5% |

#245 records those values and the reasoning in `AGENTS.md`, next to the coverage ratcheting policy, so the knowledge stays in the repo even though the settings can't.

Worth noting the original motivation has also gone away: #233 was red at "-0.3%", but that was measured against a stale base. Rebasing it onto current main turned Coveralls green on its own, without any threshold change.

PR #241, opened 2026-08-05
refactor(pds-core): extract OAuth client-id resolution from CSS middleware

aspiers:
Split 1 of 4 from #148, which bundled four independent concerns behind ~1900 lines of enrichment tests. This is the shared dependency the others build on.

## What it does

OAuth authorize pages carry either an explicit `client_id` query parameter or only a PAR `request_uri` that has to be resolved through the provider's request manager. That resolution lived inline in the CSS injection middleware. Chooser enrichment (split 4) needs exactly the same step, so this hoists it into `packages/pds-core/src/lib/oauth-request-context.ts` rather than duplicating it at the second call site.

Resolver errors stay with the caller, so each middleware keeps its own logging and fallback behaviour.

## Behaviour change

One, deliberate: the CSS middleware's failed-resolution log moves from `warn` to `error`, and `LoggerLike` gains an `error` method. A `client_id` that cannot be resolved means the page renders unbranded, which is worth an error-level line.

Otherwise this is a pure refactor — same resolution order (explicit `client_id` first, then PAR lookup), same fallbacks.

No changeset: internal refactor with no observable behaviour change for end users, client app developers, or operators.

## Verification

`typecheck`, `lint`, `format` clean; 74 test files / 1129 tests pass.

## Note on #80

This touches the same log statement that #80 redacts `requestUri` from. Whichever lands first, the other is a one-line rebase.

## Series

1. **#241 ← this PR** — extract OAuth client-id resolution
2. #242 — gate email sign-in until handlers are installed
3. #243 — sign `epds_handle_mode` through the callback hop
4. #244 — identify generated-handle accounts by email

The four branches together are byte-identical to #148's head, verified by merging them and diffing against `7fe24bb` (the only difference is the original single changeset, split into three).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OAuth client identification for direct requests and PAR flows.
  * Failed request-URI resolution no longer interrupts request processing.
  * Error logging now avoids exposing sensitive request-URI values and records only necessary metadata.
* **Reliability**
  * Added coverage for client-ID resolution, fallback behavior, invalid requests, and error handling.

### DIFF
--- a/packages/pds-core/src/__tests__/client-css-injection.test.ts
+++ b/packages/pds-core/src/__tests__/client-css-injection.test.ts
@@ -9,7 +9,7 @@ import {
 } from '../lib/client-css-injection.js'
 
 function mockLogger() {
-  return { info: vi.fn(), warn: vi.fn(), debug: vi.fn() }
+  return { info: vi.fn(), error: vi.fn(), warn: vi.fn(), debug: vi.fn() }
 }
 
 describe('shouldInjectClientCss', () => {
--- a/packages/pds-core/src/__tests__/client-css-injection.test.ts
+++ b/packages/pds-core/src/__tests__/client-css-injection.test.ts
@@ -263,6 +263,37 @@ describe('createClientCssInjectionMiddleware', () => {
     expect(next).toHaveBeenCalledOnce()
   })
 
+  // request_uri is a short-lived bearer reference to the PAR entry, so a
+  // log line carrying its value is replayable by anyone reading the logs.
+  it('logs request_uri presence but never its value when PAR resolution fails', async () => {
+    const requestUri = 'urn:ietf:params:oauth:request_uri:secret-par-handle'
+    const logger = mockLogger()
+    const middleware = createClientCssInjectionMiddleware({
+      trustedClients: [trustedClient],
+      resolveClientMetadata: vi.fn(),
+      getClientCss: vi.fn(),
+      resolveClientIdFromRequestUri: vi
+        .fn()
+        .mockRejectedValue(new Error('boom')),
+      logger,
+    })
+    const req = {
+      method: 'GET',
+      path: '/oauth/authorize',
+      query: { request_uri: requestUri },
+    }
+    const { res } = createResponseDouble()
+    const next = vi.fn()
+
+    await middleware(req, res, next)
+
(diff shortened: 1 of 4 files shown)


aspiers:
Fixed in `ee1ec80`. Both @coderabbitai and @copilot-pull-request-reviewer flagged this independently, and it's valid — the value was reaching the logs.

```diff
-        { err, requestUri: query.request_uri },
+        { err, hasRequestUri: typeof query.request_uri === 'string' },
```

`typeof` rather than the hardcoded `true` that #80 uses: this refactor hoisted the resolution out of the `if (requestUri)` guard, so reaching the catch block no longer implies a `request_uri` was present. Logging `true` unconditionally here would be wrong for `client_id` flows that fail for other reasons.

Worth being clear that I introduced the more visible half of this: the value was already being logged before this series, but promoting the line from `warn` to `error` made it more likely to be retained and shipped to log aggregation. The leak predates the split; the promotion didn't help.

Added a regression test that asserts the log context contains `hasRequestUri: true` and that the serialized context does not contain the URI value. Verified it fails against the unredacted version (1 failed / 21 passed) and passes with the fix.

#80 fixes the same leak on the pre-refactor code. Whichever lands first, the other is a small rebase.

aspiers:
Rebased onto `main` after #80 merged. That conflict was expected — both PRs changed the same log statement.

Resolved in favour of this PR's refactored form, which preserves #80's intent:

```ts
logger.error(
  { err, hasRequestUri: typeof query.request_uri === 'string' },
  'CSS middleware: failed to resolve client_id from request_uri',
)
```

#80 landed `hasRequestUri: true` inside the `if (requestUri)` guard, where presence is guaranteed. This PR hoists the resolution into `resolveOAuthClientIdFromQuery`, so the catch block is now reachable for `client_id` flows that fail for other reasons — a hardcoded `true` would be wrong there. The `typeof` form is the same redaction under the wider scope.

The `warn` → `error` promotion is unchanged from before the rebase.

Verified the full stack (#241 → #243 → #244, plus #242) still reproduces #148 exactly: merging all four branches and diffing against #148's head merged into current `main` yields only the two redaction fixes and their tests. Everything else is identical.

All four rebased branches pass typecheck, lint, format and the full vitest suite locally.

